PR #1, opened 2022-03-10
[S-RUNNER] Initial commit of Shark Runner Inferene API.

pashu123:
Added shark-runner inference API, fully functional given only one input.


dan-garvey:
Missed the review request. I have also never built a pip package, but as long as we follow the general structure of organizing stuff into folders based on functionality, we should be fine

PR #11, opened 2022-03-25
Add scripts to generate Bert graph and run it on IREE

ramiro050:
This commit adds two scripts. One generates the graph for the forward
pass of Bert training at the MLIR linalg+tensor dialect level. The
other script can take any linalg+tensor graph and run it on IREE.


powderluv:
@pashu123 @ramiro050 is this still the right way to generate the BERT training graph ?

PR #14, opened 2022-03-31
Update readme.md of shark_runner

pashu123:
Updated the readme.md of shark_runner and added setup.py for pip
installation.


pashu123:
@powderluv I have updated the README. PTAL.

PR #36, opened 2022-05-05
Add function for models benchmarking

vivekkhandelwal1:


shark/shark_runner.py
```diff
@@ -84,12 +85,25 @@ def __init__(
...
         # forward pass.
+        begin = time.time()
         inputs = self.input if self.from_aot else inputs
         input_list = [x.detach().numpy() for x in inputs]
-        return self.shark_runner.forward(input_list)
+        out = self.shark_runner.forward(input_list)
+        end = time.time()
+        return out, end - begin
+
+    def forward(self, inputs):
```
pashu123, line 98: The forward function shouldn't change or have any timings. Just create debug_forward same as forward with timings.

pashu123 (changes requested):

shark/shark_runner.py
powderluv: benchmark_forward instead of debug_forward

vivekkhandelwal1: And the models by default will call benchmark_forward?

powderluv: yeah since they are samples that is ok.  We can add a comment saying switch to forward() if using in production.

shark/examples/test_cpu.py
```diff
@@ -70,7 +70,7 @@ def test_bert():
     shark_module = SharkInference(
         model, (input,), device="cpu", jit_trace=True
     )
-    results = shark_module.forward((input,))
+    results = shark_module.benchmark_forward((input,))
```
pashu123, line 73: In `test_cpu.py` please use forward instead of benchmark_forward since we this will be running in the CI and hence can be slow.

pashu123 (commented):
Only one NIT, rest LGTM.

PR #51, opened 2022-05-21
Add support for TensorFlow modules + TF miniLM example

monorimet:
MiniLM example sourced from iree-samples, modified to run through SHARK.

### DIFF
--- a/shark/shark_runner_tf.py
+++ b/shark/shark_runner_tf.py
@@ -12,12 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from shark.iree_utils import get_results
-from shark.iree_utils_tf import get_iree_compiled_module_tf, export_tf_iree_module_to_vmfb
+from shark.iree_utils_tf import get_iree_compiled_module_tf, export_tf_iree_module_to_vmfb, get_results_tf
 import os
-from shark.functorch_utils import AOTModule
+import numpy as np
 from shark.parser import shark_args
-from shark.backward_makefx import MakeFxModule
 from tqdm import tqdm
 import time
 
@@ -29,10 +27,7 @@ class SharkRunnerTF:
         self,
         model,
         input: tuple,
-        dynamic: bool,
         device: str,
-        tracing_required: bool,
-        from_aot: bool,
     ):
         self.tf_module = model
         self.input = input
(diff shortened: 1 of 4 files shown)


shark/iree_utils.py
```diff
@@ -14,6 +14,7 @@
 
 import iree.runtime as ireert
 import iree.compiler as ireec
+from iree.compiler import tf as tfc
```
powderluv, line 17: Can we move this to iree_tf_utils.py ? We want to be able to run torch inference even if tensorflow and IREE tf is not installed.

pashu123:
I don't know whether passing frontend as an arg is a good idea. Instead, we can have `shark.set_frontend("frontend")` and have shark_runner copies for that frontend, this will decouple the behaviour and remove if/else's.

powderluv:
true. maybe we model it on the HF trainer apis: the Default can be like the HF trainer: https://github.com/huggingface/transformers/blob/main/src/transformers/trainer.py

and the TF version is: 
https://github.com/huggingface/transformers/blob/main/src/transformers/trainer_tf.py 

So we don't have a bunch of if/else but just two entry points. Default one can be Torch.

shark/shark_runner_tf.py
```diff
@@ -0,0 +1,151 @@
...
+
+        self.weights = [
+            i[1] for i in sorted(dict(model.named_parameters()).items())
+        ]
+        for i in sorted(dict(model.named_buffers()).items()):
+            self.weights.append(i[1])
+
+        for i in input:
+            self.input.append(i)
+
```
pashu123, line 137: I think `model.named_parameters()` is a PyTorch API call and TensorFlow would have a different call. This should change.

monorimet:
I managed to get both included examples running as expected -- was running into issues with the IREE Device Arrays; copying the get_results function from iree_utils.py into a get_results_tf function that skips the input conversion to IREE DeviceArrays solved the problem (inputs get compiled with the tf.Module when generating MLIR via IREE's tfc.compile_module, so input conversion to device arrays for TF doesn't seem necessary --  please correct me if I'm wrong.)

powderluv:
You can do device array in a follow on commit

powderluv (approved):

PR #62, opened 2022-05-26
Add Shark Benchmark

raikonenfnu:
-Introduce SharkBenchmark that bench models on regular torch, shark-py, and shark-c.
-Integrate iree-benchmark-module into Shark.
-Refactor SharkRunner to have low level API control.

### DIFF
--- a/shark/iree_utils.py
+++ b/shark/iree_utils.py
@@ -13,11 +13,14 @@
 # limitations under the License.
 
 import iree.runtime as ireert
+import iree.runtime.scripts.iree_benchmark_module as benchmark_module
 import iree.compiler as ireec
 from iree.compiler import tf as tfc
+from shark.torch_mlir_utils import get_module_name_for_asm_dump
 import subprocess
 import numpy as np
 import os
+import re
 
 IREE_DEVICE_MAP = {
     "cpu": "dylib",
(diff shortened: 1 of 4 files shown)


powderluv:
Push and Iterate.

pashu123:
One question: Why can't benchmark be a function under SharkInference? Or just a SharkBenchMarkRunner just like SharkRunner that can be called inside SharkInference - so instead of forward we can call the benchmark function.

raikonenfnu:
> One question: Why can't benchmark be a function under SharkInference? Or just a SharkBenchMarkRunner just like SharkRunner that can be called inside SharkInference - so instead of forward we can call the benchmark function.

Integrated it into ShakRunner :)

powderluv (approved):

shark/iree_utils.py
```diff
@@ -152,6 +159,15 @@ def get_iree_compiled_module(module,
 
     return get_iree_module(module, device, input_type, args, func_name)
 
+def export_iree_module_to_vmfb(module, device: str, directory: str):
```
powderluv, line 162: Nit:Shark module? In our code we can call IREE SHARK

powderluv (approved):

shark/iree_utils.py
raikonenfnu: There's actually a bunch of these in iree_utils, I'll push this up in a separate patch! :)

PR #63, opened 2022-05-26
Refactor the whole SharkInference API.

pashu123:
1. SharkInference API is refactored to target any frontends or files,
   i.e. pytorch, tensorflow, mhlo, linalg, tosa.
2. Refactor all torch APIs.


powderluv (approved):

raikonenfnu:
@pashu123 LGTM, but I have some questions. What is the rationale behind moving the compile step into SharkInference, and why did we remove the save_mlir and save_vmfb, which is quite useful to debug the code generated.

powderluv:
I think we could do a low level api:  compile / run  and a high level: SharkInference()

save mlir / vmfb seems like we need to re-add those options.

powderluv:
@raikonenfnu now it is your turn to push and iterate. so when @pashu123 wakes up it has evolved :D

PR #66, opened 2022-05-27
Pass --iree-hal-cuda-llvm-target-arch for GPU execution automatically.

monorimet:
Also adds in IREE DeviceArray support for Tensorflow frontend.


powderluv:
Commit message needs to be updated

powderluv (commented):
Let's set the flag only if sm 70 or above

pashu123:
I have updated the shark_trainer. PTAL.

pashu123:
Also, @monorimet If you can modify this example: https://github.com/nod-ai/SHARK/blob/main/shark/examples/shark_training/bert_training_tf.py That would be great.

PR #78, opened 2022-06-01
TorchDynamo demo

makslevental:
Demo for how dynamo with mlir backend can exist out of (dynamo) tree. WIP depending on what we want to show off here.

cc @powderluv

### DIFF
--- a/shark/examples/shark_eager/dynamo_demo.py
+++ b/shark/examples/shark_eager/dynamo_demo.py
@@ -1,0 +1,79 @@
+import torch
+import torchdynamo
+from torch_mlir import compile, OutputType
+from torchdynamo.optimizations.backends import create_backend
+from torchdynamo.optimizations.subgraph import SubGraph
+
+from shark.iree_utils import get_iree_compiled_module
+
+NUM_ITERS = 10
+
+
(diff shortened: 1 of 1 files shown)


shark/examples/shark_eager/dynamo_demo.py
```diff
@@ -0,0 +1,79 @@
...
+        return fx_g
+
+    fx_graph = _unwrap_single_tuple_return(fx_graph)
+    ts_graph = torch.jit.script(fx_graph)
+
+    if isinstance(args, tuple):
+        args = list(args)
+    assert isinstance(args, list)
+    if len(args) == 1 and isinstance(args[0], list):
+        args = args[0]
```
makslevental, line 37: dynamo does some strange things with args depending on how you use it

powderluv:
Nice. Can we add a notebook version of this too ? 
Do we need to pip install torchdynamo ? or is already part of the nightly ?

makslevental:
>Nice. Can we add a notebook version of this too ?

sure coming right up

>Do we need to pip install torchdynamo ? or is already part of the nightly ?

not part of nightly yet so we'll have to add as a dep

PR #80, opened 2022-06-01
Add support for training TF models + fix TF BERT training example

monorimet:


shark/shark_runner.py
```diff
@@ -48,10 +48,16 @@ def __init__(
...
             self.iree_compilation_module,
             self.iree_config,
-        ) = get_iree_compiled_module(self.model, device)
+            ) = get_iree_compiled_module(self.model, device)
+
+        if self.frontend in ["tensorflow", "tf", "mhlo"]:
+            (
+            self.iree_compilation_module,
+            self.iree_config,
+            ) = get_iree_compiled_module(self.model, device, self.frontend)
```
raikonenfnu, line 60: Can be combined into just get_iree_compiled_module(self.model, device, self.frontend)

shark/iree_utils.py
```diff
@@ -135,12 +135,18 @@ def get_iree_frontend_args(frontend):
...
             str(module),
             target_backends=[IREE_DEVICE_MAP[device]],
             extra_args=args,
             input_type=input_type)
+    elif input_type in ["mhlo"]:
+        flatbuffer_blob = ireec.compile_str(
+            module,
+            target_backends=[IREE_DEVICE_MAP[device]],
+            extra_args=args,
+            input_type=input_type)
```
raikonenfnu, line 149: can we just add if input_type = tosa: module = str(module)?

raikonenfnu: Actually, it's fine for now, I need to also push up a change that has conflict with this, will clean up there :)

shark/shark_runner.py
raikonenfnu: Actually, it's fine for now, I need to also push up a change that has conflict with this, will clean up there :)

shark/iree_utils.py
powderluv: +2 for push and iterate

PR #82, opened 2022-06-02
Refactor get_iree_module to work with SharkBenchmark

raikonenfnu:
1. Refactored get_iree_module to a function that convert mlir->vmfb. (helps with generation of correct vmfb for benchmarking)
2. Added benchmarking test
3. minor cleanups and fix

Example:

```bash
python -m shark.examples.shark_inference.minilm_benchmark --device="cpu"
```

### DIFF
--- a/shark/iree_utils.py
+++ b/shark/iree_utils.py
@@ -131,28 +131,38 @@ def get_iree_frontend_args(frontend):
         return []
 
 
-# input_type should be "mhlo", "tosa" for linalg no need to mention the frontend.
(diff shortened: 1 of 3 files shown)


powderluv:
thank you for the tests.

raikonenfnu:
> thank you for the tests.

np! would be good to track that benchmarking is working for the different frontends and backend. which I will add in the future. May also be good for performance/speed tracking/checking. :)

powderluv (approved):

PR #83, opened 2022-06-02
Add iree-tools-tflite iree-tools-tf tensorflow in python venv

AmosLewis:


powderluv (commented):
I have a change for this. Since this has to run on OSX too we don't have those tools in osx. I will push up my change in a few mins.

PR #87, opened 2022-06-04
Add benchmark for TF

raikonenfnu:
-Refactor SharkBenchmarker to run TF
-Add example and test to benchmark TF

### DIFF
--- a/setup_venv.sh
+++ b/setup_venv.sh
@@ -43,7 +43,9 @@ Yellow=`tput setaf 3`
 torch_mlir_bin=false
 if [[ $(uname -s) = 'Darwin' ]]; then
   echo "${Yellow}Apple macOS detected"
+  install_tensorflow_mac=true
   if [[ $(uname -m) == 'arm64' ]]; then
+    install_tensorflow_metal_extension=true
     echo "${Yellow}Apple M1 Detected"
     hash rustc 2>/dev/null
     if [ $? -eq 0 ];then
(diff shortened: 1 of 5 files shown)


setup_venv.sh
```diff
@@ -64,6 +65,7 @@ elif [[ $(uname -s) = 'Linux' ]]; then
   echo "${Yellow}Linux detected"
   if [ "$PYTHON_VERSION_X_Y" == "3.7" ] || [ "$PYTHON_VERSION_X_Y" == "3.8" ]  || [ "$PYTHON_VERSION_X_Y" == "3.9" ] || [ "$PYTHON_VERSION_X_Y" == "3.10" ] ; then
     torch_mlir_bin=true
+    install_tensorflow=true
```
powderluv, line 68: I think requirements-importer.txt should already bring in tf no ?

setup_venv.sh
```diff
@@ -72,6 +74,17 @@ fi
 # Upgrade pip and install requirements.
 $PYTHON -m pip install --upgrade pip || die "Could not upgrade pip"
 $PYTHON -m pip install --upgrade -r "$TD/requirements.txt" --extra-index-url https://download.pytorch.org/whl/nightly/cpu -f https://github.com/llvm/torch-mlir/releases
+if [ "$install_tensorflow" = true ]; then
+  $PYTHON -m pip install tensorflow
+  $PYTHON -m pip install --find-links https://github.com/google/iree/releases iree-tools-tf
```
powderluv, line 79: this should already happen on linux

raikonenfnu: got it!

setup_venv.sh
raikonenfnu: yeap, you're right

raikonenfnu: done!

setup_venv.sh
raikonenfnu: done!

powderluv (approved):

PR #89, opened 2022-06-05
Add saving and loading of mlir to run.

raikonenfnu:
Added:
-Generalized save_mlir for tensorflow+torch
-Return loss for tf train since that's what we need to observer training 
-Remove return of self.trainable_params isn't updated in the mlir runs.

To try load and save:
```bash
# Saves mlir file to /tmp/model.mlir
python -m shark.examples.shark_training.bert_training_load_tf --device="cpu" --save_mlir
# Runs mlir file in /tmp/model.mlir
python -m shark.examples.shark_training.bert_training_load_tf --device="cpu"
# If you want to download in a specific directory and filename
python -m shark.examples.shark_training.bert_training_load_tf --device="cpu" --download_mlir_path=/tmp/tf_bert.mlir
```
Should work on device of your choice(cpu/gpu/vulkan).

### DIFF
--- a/shark/examples/shark_training/bert_training_load_tf.py
+++ b/shark/examples/shark_training/bert_training_load_tf.py
@@ -1,0 +1,49 @@
+import numpy as np
+import os
+import time
+import tensorflow as tf
+
+from shark.shark_trainer import SharkTrainer
+from shark.parser import parser
+from urllib import request
+
+parser.add_argument(
+    "--download_mlir_path",
+    type=str,
+    default="bert_tf_training.mlir",
+    help="Specifies path to target mlir file that will be loaded.")
+load_args, unknown = parser.parse_known_args()
+
+tf.random.set_seed(0)
+vocab_size = 100
+NUM_CLASSES = 5
+SEQUENCE_LENGTH = 512
+BATCH_SIZE = 1
+
+# Download BERT model from tank and train.
+if __name__ == "__main__":
+    predict_sample_input = [
+        np.random.randint(5, size=(BATCH_SIZE, SEQUENCE_LENGTH)),
+        np.random.randint(5, size=(BATCH_SIZE, SEQUENCE_LENGTH)),
+        np.random.randint(5, size=(BATCH_SIZE, SEQUENCE_LENGTH))
+    ]
+    file_link = "https://storage.googleapis.com/shark_tank/users/stanley/bert_tf_training.mlir"
+    response = request.urlretrieve(file_link, load_args.download_mlir_path)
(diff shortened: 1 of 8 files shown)


shark/examples/shark_training/bert_training_load_tf.py
```diff
@@ -0,0 +1,44 @@
...
+
+if __name__ == "__main__":
+    predict_sample_input = [
+        np.random.randint(5, size=(BATCH_SIZE, SEQUENCE_LENGTH)),
+        np.random.randint(5, size=(BATCH_SIZE, SEQUENCE_LENGTH)),
+        np.random.randint(5, size=(BATCH_SIZE, SEQUENCE_LENGTH))
+    ]
+    sample_input_tensors = [tf.convert_to_tensor(val, dtype=tf.int32) for val in predict_sample_input]
+    num_iter = 10
+    mlir_path = "/tmp/model.mlir"
```
powderluv, line 30: can we please default to $PWD/filename.mlir ?  (When running on builders etc we may not have access to /tmp/)

shark/examples/shark_training/bert_training_load_tf.py
```diff
@@ -0,0 +1,51 @@
...
+
+from official.nlp.modeling import layers
+from official.nlp.modeling import networks
+from official.nlp.modeling.models import bert_classifier
+
+from shark.shark_trainer import SharkTrainer
+from shark.parser import parser
+
+parser.add_argument(
+    "--target_mlir",
```
powderluv, line 18: nit. load_mlir may be better to match save_mlir but that can be changed in a future commit.

powderluv (approved):

shark/examples/shark_training/bert_training_load_tf.py
raikonenfnu: done! for now you can actually target where do you want to download the file into and the code will look for it. Of course the generalization of this example is to have a loader that can given a local mlir file you can load it in without downloading ahead of time.

shark/examples/shark_training/bert_training_load_tf.py
raikonenfnu: haha, true, it was initially assuming you have a local mlir file. I can change it to download_mlir_path, so it won't conincide with the save_mlir flag which you need to initially save the mlir.

powderluv: or just mlir_path (since you may not always download it : )

powderluv:
feel free to merge when ready

PR #92, opened 2022-06-05
fixed issues now that shark-runtime patch is added

Eliasj42:


.gitmodules
```diff
@@ -1,3 +1,4 @@
 [submodule "inference/thirdparty/shark-runtime"]
 	path = inference/thirdparty/shark-runtime
-	url = https://github.com/NodLabs/shark-runtime.git
+	url =https://github.com/nod-ai/SHARK-Runtime.git
+	branch = shark-06032022
```
powderluv, line 4: i think you don't need the branch here since you specify the SHA later. But we can clean up in the next patch.

inference/README.md
```diff
@@ -4,26 +4,25 @@ The triton backend for shark.
 
 # Build
 
-Install dSHARK
+Install SHARK
```
powderluv, line 7: I made these edits this morning. Just make sure you are on top of main

inference/src/dshark.cc
```diff
@@ -1238,7 +1237,7 @@ ModelInstanceState::Execute(
     IREE_LOG(INFO) << "can't find return buffer view";
   }
 
-  IREE_CHECK_OK(iree::PrintVariantList(output_tensors, (size_t)1024));
+  //IREE_CHECK_OK(iree::PrintVariantList(output_tensors, (size_t)1024));
```
powderluv, line 1240: remove debugs

Eliasj42: I deleted this line

inference/README.md
Eliasj42: It isn't reporting any merge conflicts though? is it fine to just merge it?

.gitmodules
Eliasj42: don't you need the branch because thats where I pushed the api updates to?  without it, won't it pull main, which doesn't have the function

PR #94, opened 2022-06-06
Add distilbert test case.

pashu123:


powderluv (approved):

pashu123:
> this is ok to merge. Once the new pytest framework comes in we can move into tank/pytorch/

Sounds good.

PR #96, opened 2022-06-06
Add torchvision as a dependency to resent50 sample

sogartar:


powderluv:
just saw this PR. Rolled a version of this already. thanks for the fix and testing.

PR #113, opened 2022-06-09
Refactor testing framework.

monorimet:


powderluv:
Let's merge and get it fixed in follow-on commits. Please get @raikonenfnu  to review

powderluv (commented):
Can we please move everything under the new models/ directory into the canonical tank/torch/ or tank/tf ? 

Also can we please name the model files as MODELNAME_test.py ? (pytest should still pickup) but inference can be implied and we can have a MODELNAME_training.py later. 

The new SHARK build renamed `dylib` to `local-task` so we need to update that. 

Lastly can we please add a __main__ so each model can be run as is without Pytest. See https://github.com/nod-ai/SHARK/blob/main/tank/tflite/albert.py. 

We will need to reconcile and update the TFlite tests once yours lands.

powderluv:
Also for macOS if it is just torchvision and tqdm please readd and re-enable the builder.

PR #117, opened 2022-06-13
Refactor testing framework.

monorimet:
- Separate model tests into individual python scripts formatted as `{MODELNAME}_test.py`
- Remove pytest parametrization of tests and replace with unittest framework.
- Move common testing utility classes/functions to test_utils.py (test_utils_tf.py).
- Move all model tests to SHARK/tank/torch/tests/ or SHARK/tank/tf/tests/.
- Move benchmark tests to SHARK/benchmarks/tests/.
- Enable each test to be run independently of pytest (via `python -m unittest tank.torch.tests.bert_test.BertModuleTest.test_module_static_cpu` , for example).

In upcoming PR:
- Modify benchmark tests to match this testing framework.
- Add in functional TensorFlow model tests.


powderluv:
Merged. Let's iterate from here.

Issue #120, opened 2022-06-13
Use TEmpFileSaver API in IREE


powderluv:
Use TempFileSAver to save each PyTest output artifacts and recreate if the test fails. 

looks like you can also write your own TempFileSaver to do whatever you want (see the section just below that), like in this test: https://github.com/google/iree/blob/427a94a09be70631c5d9f89d12616bb7f1954257/compiler/src/iree/compiler/API/python/test/tools/compiler_core_test.py#L176-L199

https://discord.com/channels/689900678990135345/689900680009482386/985726147448950805

monorimet:
#149 
Just need to get these tests refactored for importer/downloader

monorimet:
Is this still necessary now that we have generate_sharktank? Perhaps we can add options to generate_sharktank to save the command line string for iree-compile if we want equivalent functionalities between what we have and what IREE TempFileSaver gives us.

PR #122, opened 2022-06-14
Enable passing of --save-mlir cmd-line option through pytest.

monorimet:
Save MLIR input for individual test cases by passing `--save-mlir` option through pytest. Can set `IREE_SAVE_TEMPS=<dir>` env variable to specify directory to which MLIR input will be exported.

Example:
Running from base SHARK directory,

`IREE_SAVE_TEMPS=iree_temps_bert_dynamic  pytest --save_mlir tank/pytorch/bert_test.py::BertModuleTest::test_module_dynamic_cpu`

will save the input MLIR to `SHARK/iree_temps_bert_dynamic/core-input.mlir`.

If a directory is not specified, the MLIR input will be saved to `/tmp/model.mlir` by default.

Note: currently, names for saved MLIR input files are not unique, so it is not advised to use the --save_mlir option with pytest when running more than one test case at a time.

### DIFF
--- a/tank/conftest.py
+++ b/tank/conftest.py
@@ -1,3 +1,3 @@
 def pytest_addoption(parser):
     # Attaches SHARK command-line arguments to the pytest machinery.
-    parser.addoption("--save_mlir", action="store_true", help="Pass option to save input MLIR module to /tmp/ directory.")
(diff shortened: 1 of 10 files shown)


powderluv:
I have https://github.com/nod-ai/SHARK/issues/120 for naming the files so we can run them all together.

powderluv (approved):

PR #124, opened 2022-06-14
Add more tflite examples for shart_importer

AmosLewis:
Add all the available tflite examples that use shark_importer. Some of them have bugs which are from the original iree/samples. Most of the bug is because iree-tflite-tools do not support dynamic inputs right now. I leave the bug output at the end of those files.


raikonenfnu:
@AmosLewis can you check why the tests are failing?

powderluv:
still using old iree calls 

``` 
RuntimeError: Could not create any requested driver ['dylib'] (available=['cuda', 'local-sync', 'local-task', 'vulkan']) : {}
```

raikonenfnu:
Ahhh, I think he was still using an older venv when pushing up these changes "local-sync', 'local-task" did not exist when he was trying it out on his local machine.

PR #127, opened 2022-06-14
Add seq_classification shark module to run seq_classification

pashu123:
on hugging face supported models.


pashu123:
Also, let's try to remove the ALBERT test from CI until it's green all the way.

PR #128, opened 2022-06-14
Enable Benchmark Tests

raikonenfnu:


powderluv:
you may want to disable it on the "Github builder"

.github/workflows/test-models.yml
```diff
@@ -52,7 +52,7 @@ jobs:
         cd $GITHUB_WORKSPACE
         IMPORTER=1 ./setup_venv.sh
         source shark.venv/bin/activate
-        pytest -k 'not benchmark' --ignore=tank/tf/ --ignore=shark/tests/test_shark_importer.py
+        pytest --ignore=tank/tf/ --ignore=shark/tests/test_shark_importer.py
```
powderluv, line 55: i would remove it in the perf-linux and perf-macOS bots and leave it disabled in the GH builder

raikonenfnu:
@powderluv it's weird tho, seems to be passing for all the benchmark tests, but break at the very end:
https://github.com/nod-ai/SHARK/runs/6885863019?check_suite_focus=true#step:7:465

powderluv:
not worth benchmarking on a dual core VM instances that is over subscribed. It probably kills itself with an OOM or something

powderluv:
i pushed a commit when updating the macOS builder.

PR #130, opened 2022-06-15
Mini LM Loader Example

raikonenfnu:


shark/examples/shark_inference/minilm_load_benchmark_tf.py
```diff
@@ -0,0 +1,45 @@
...
+    text = "Replace me by any text you'd like."
+    encoded_input = tokenizer(text,
+                              padding='max_length',
+                              truncation=True,
+                              max_length=MAX_SEQUENCE_LENGTH)
+    for key in encoded_input:
+        encoded_input[key] = tf.expand_dims(
+            tf.convert_to_tensor(encoded_input[key]), 0)
+    file_link = "https://storage.googleapis.com/shark_tank/users/stanley/minilm_tf_inference.mlir"
+    response = request.urlretrieve(file_link, load_args.download_mlir_path)
```
powderluv, line 34: this can be wrapped into a share_loader() class and the URL can be shortened to only require the model name.

raikonenfnu: done!

powderluv (approved):

powderluv:
Lgtm

shark/shark_importer.py
```diff
@@ -122,3 +122,15 @@ def forward(self, inputs=None):
             print("Inference", self.model_source_hub, " not implemented yet")
         elif self.model_source_hub == "jaxhub":
             print("Inference", self.model_source_hub, " not implemented yet")
+
+
+def shark_load(model_name, file_path):
```
pashu123, line 127: The name should be something that specifies loading from shark_hub like `load_from_shark_hub()` or `shark_hub_load()`.

shark/shark_importer.py
```diff
@@ -122,3 +122,15 @@ def forward(self, inputs=None):
...
+    file_link = f"https://storage.googleapis.com/shark_tank/users/stanley/{model_name}.mlir"
```
pashu123, line 128: Can't this be generic file_link, instead of specifying user_name etc.

shark/examples/shark_training/bert_training_load_tf.py
```diff
@@ -27,14 +27,10 @@
         np.random.randint(5, size=(BATCH_SIZE, SEQUENCE_LENGTH)),
         np.random.randint(5, size=(BATCH_SIZE, SEQUENCE_LENGTH))
     ]
-    file_link = "https://storage.googleapis.com/shark_tank/users/stanley/bert_tf_training.mlir"
-    response = request.urlretrieve(file_link, load_args.download_mlir_path)
+    model_name = "bert_tf_training"
```
pashu123, line 30: I think the model name should be equivalent to the hugging face model name with {training} and {inference} at the end. Users will find it easy to run the model.

shark/examples/shark_inference/minilm_load_benchmark_tf.py
```diff
@@ -0,0 +1,41 @@
...
+import os
+
+gpus = tf.config.experimental.list_physical_devices('GPU')
+for gpu in gpus:
+  tf.config.experimental.set_memory_growth(gpu, True)
+
+parser.add_argument(
+    "--download_mlir_path",
+    type=str,
+    default="minilm_tf_inference.mlir",
```
pashu123, line 15: something like `microsoft/MiniLM-L12-H384-uncased_inference`, we can later add the .mlir extension. This makes sure that user can navigate to and fro shark_hub and hugging face to find the right model.

pashu123 (changes requested):
Some comments. Rest LGTM.

shark/examples/shark_training/bert_training_load_tf.py
powderluv: agreed. I would even drop _inference at the end and assume inference by default. Beside for training it could just be a flag to shark_load() to get the training graph and what framework etc.

shark/shark_importer.py
powderluv: I like shark_load() to keep it small and simple but I am not sure if we want to call hub / zoo yet (maybe tank ?) but this is a marketing exercise :D

shark/shark_importer.py
powderluv: I suggested users/ for custom user generated models. The official models once they are loaded into shark_tank/<model_name> should be simple like HF namespace.

shark/examples/shark_training/bert_training_load_tf.py
raikonenfnu: this is not a huggingface model per se, but we do need a nice way to name it :)

PR #132, opened 2022-06-15
Increase IREE LLVM stack alloc limit for dynamic cpu case on Language Models.

monorimet:
Adds a flag to shark parser that increases LLVMCPU stack allocation size limit from 32KB to 128KB.
This fixes issues with running the dynamic case for language models on CPU.


powderluv:
lets hold on this commit. They merged a patch and I have started a build. So in 2 hrs we can see how they work

PR #134, opened 2022-06-16
Fixed bug where backend would segfault under specific circumstances

Eliasj42:
backend would sometimes segfault if a model had multiple outputs of different sizes when run on gpu


inference/src/dshark.cc
```diff
@@ -913,7 +928,7 @@ ModelInstanceState::ProcessRequests(
 
 
   if (!all_response_failed) {
-    Execute(&responses, requests, request_count, request_count, input_tensors, output_tensors, output_names);
+    Execute(&responses, requests, request_count, request_count, input_tensors, output_tensors, output_names, output_dtypes, output_dims);
```
pashu123, line 931: Please use clang-format.

powderluv:
is this ready to merge ?

Eliasj42:
> 

yeah

PR #143, opened 2022-06-20
Add TempFileSaver option to tf tests and fix commit history.

monorimet:
Adds the option to save reproduction artifacts for TF tests in tank.

To save reproduction artifacts from a test run, add the --save_temps option to your pytest command.

Example:
Pytest invocation:
Results:


monorimet:
Got rid of my changes somehow. Fixing it

PR #145, opened 2022-06-21
Enable tosa.mlir as input for SharkImporter inference

AmosLewis:
In this patch, we use the tosal.mlir file in local gen_shark_tank folder to do the inference in SharkImporter.
I also create a new albert_lite_base_tflite_test.py in tank/tflite folder to replace the old albert.py. This test is the same format as tank/tf and tank/pytorch folder.


powderluv:
merging so @pashu123 can build on it

PR #147, opened 2022-06-21
Add SharkDownloader for end users

AmosLewis:
I add a shark_downloader.py for end users. Uses should give model name, model type, input type, input json, and local shark tank dir. If the model is not found locally, the user will have to give tank_url for download. Right now, it only supports the tosa.mlir file converts from tflite mode. Will add other support later.
The albert_lite_base_tflite_mlir_test.py show you how to use it. Here is quick reference:
        self.shark_downloader = SharkDownloader(model_name="albert_lite_base",
                                                tank_url="https://storage.googleapis.com/shark_tank/tflite"
                                                         "/albert_lite_base/albert_lite_base_tosa.mlir",
                                                local_tank_dir="./../gen_shark_tank/tflite",
                                                model_type="tflite-tosa",
                                                input_json="input.json",
                                                input_type="int32")
        tflite_tosa_model = self.shark_downloader.get_mlir_file()
        inputs = self.shark_downloader.get_inputs()
        self.shark_module = SharkInference(tflite_tosa_model,
                                           inputs,
                                           device=self.device,
                                           dynamic=self.dynamic,
                                           jit_trace=True)
        self.shark_module.set_frontend("tflite-tosa")
        self.shark_module.compile()
        self.shark_module.forward(inputs)


shark/shark_downloader.py
```diff
@@ -0,0 +1,145 @@
...
+            else:
+                print("Download mlir model")
+                urllib.request.urlretrieve(self.tank_url,
+                                           self.tflite_tosa_mlir_file)
+
+    def setup_inputs(self, inputs):
+        print("Setting up inputs. Input has to be (list of np.array)")
+        self.inputs = inputs
+
+    def compile(self, inputs=None):
```
powderluv, line 109: why do we have compile / run in the downloader ?

AmosLewis: Did you want me to make the compile and run just in Sharkimporter, or delete compile and forward in Sharkdownloader. But the sharkimporter now it try to deal with the .tflite file. We want a Download to just dealing with the mile file and run with our sharkinference.  Or we can leave the end user directly call sharkinference after sharkdownloader in test file by themselves?

tank/tflite/albert_lite_base/albert_lite_base_tflite_mlir_test.py
```diff
@@ -0,0 +1,83 @@
...
+        self.dynamic = dynamic
+        self.device = device
+        self.save_mlir = save_mlir
+        self.save_vmfb = save_vmfb
+
+    def create_and_check_module(self):
+        shark_args.save_mlir = self.save_mlir
+        shark_args.save_vmfb = self.save_vmfb
+        self.shark_downloader = SharkDownloader(model_name="albert_lite_base",
+                                                tank_url="https://storage.googleapis.com/shark_tank/tflite"
```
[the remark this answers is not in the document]
AmosLewis: Done

raikonenfnu:
Code looks good, but for the test are we not going to test or compare the output for correctness? or should we add this tests later? @powderluv

AmosLewis:
> Code looks good, but for the test are we not going to test or compare the output for correctness? or should we add this tests later? @powderluv

I think our goal is to make end-user avoid using any of tf/torch packages? If we want to compare the result, it will import them again.

raikonenfnu:
> > Code looks good, but for the test are we not going to test or compare the output for correctness? or should we add this tests later? @powderluv
> 
> I think our goal is to make end-user avoid using any of tf/torch packages? If we want to compare the result, it will import them again.

I think we had the idea of downloading npz from the tank, S.T we can load to check the input and expected output.

AmosLewis:
> 

Then we would be better to change the generate_tank to get the npz etc..... Then we come back to change the sharkdownloader.

raikonenfnu:
Cool cool, in that case, looks good to me :)

PR #148, opened 2022-06-21
Equip TF tests with save_mlir, save_vmfb, and benchmark to .csv options.

monorimet:
Use --benchmark flag with pytest to generate a bench_results.csv that contains results for inference benchmarks on frontend, SHARK-python, and SHARK-iree-c for each test case in the pytest run.

This commit only adds benchmark functionality to TensorFlow tests. Support for other frontends will come in follow-up commits.

These TF tests should serve as a model for how tests should be structured in all frontends, BEFORE the importer/downloader split happens.

Example .csv obtained by running static cpu TF tests with the --benchmark option:
[bench_results.csv](https://github.com/nod-ai/SHARK/files/8953374/bench_results.2.csv)

### DIFF
--- a/shark/iree_utils.py
+++ b/shark/iree_utils.py
@@ -290,18 +290,34 @@ def tensor_to_type_str(input_tensors: tuple, frontend: str):
     """
     list_of_type = []
     for input_tensor in input_tensors:
-        type_string = "x".join([str(dim) for dim in input_tensor.shape])
-        if frontend in ["torch", "pytorch"]:
-            dtype_string = str(input_tensor.dtype).replace("torch.", "")
-        elif frontend in ["tensorflow", "tf"]:
-            dtype = input_tensor.dtype
-            dtype_string = re.findall('\'[^"]*\'',
-                                      str(dtype))[0].replace("\'", "")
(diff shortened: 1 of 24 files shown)


powderluv:
this is looking great.

shark/iree_utils.py
```diff
@@ -290,18 +290,34 @@ def tensor_to_type_str(input_tensors: tuple, frontend: str):
...
-            dtype_string = re.findall('\'[^"]*\'',
-                                      str(dtype))[0].replace("\'", "")
-        regex_split = re.compile("([a-zA-Z]+)([0-9]+)")
-        match = regex_split.match(dtype_string)
-        mlir_type_string = str(match.group(1)[0]) + str(match.group(2))
-        type_string += f"x{mlir_type_string}"
-        list_of_type.append(type_string)
+        if (isinstance(input_tensor, tuple)):
+            for val in input_tensor:
+                type_string = "x".join([str(dim) for dim in val.shape])
```
raikonenfnu, line 295: Looks like duplicate code, let's move the original function a utility function perhaps calling it something like get_tensor_mlir_type

raikonenfnu: aside from this lgtm! :)

PR #157, opened 2022-06-24
Fix repo_dir cannot export/write mlir file bug

AmosLewis:


shark/parser.py
```diff
@@ -44,8 +44,7 @@ def dir_file(path):
 parser.add_argument(
     "--repro_dir",
     help="Directory to which module files will be saved for reproduction or debugging.",
-    type=dir_path,
-    default="./repro_dir/",
+    default="./tmp/",
```
raikonenfnu, line 47: Can we change this to `./shark_tmp`

raikonenfnu, line 47: Can we then add `./shark_tmp` into `.gitignore`?

AmosLewis: done

AmosLewis: How

monorimet:
The IREE TempFileSaver doesn't call the mlir/vmfb export functions -- can we put the directory creation in shark/parser.py when it checks the repro_dir path? Otherwise the shark_tmp directory won't be created when only using the --save_temps flag.

AmosLewis:
> The IREE TempFileSaver doesn't call the mlir/vmfb export functions -- can we put the directory creation in shark/parser.py when it checks the repro_dir path? Otherwise the shark_tmp directory won't be created when only using the --save_temps flag.

Done

raikonenfnu (approved):

PR #159, opened 2022-06-24
SharkImporter for tflite without forward and compile

AmosLewis:
Delete SharkImporter forward and compile. Make it return mlir_model so SharkInference can use it as the input model.
Rewrite the tflite SharkImporter test file: albert_lite_base_tflite_test.py
Quick reference on how to use it:
```
my_shark_importer = SharkImporter(
    model_name="albert_lite_base",
    model_type="tflite"
)
mlir_model = my_shark_importer.get_mlir_model()
inputs = my_shark_importer.get_inputs()
shark_module = SharkInference(
    mlir_model,
    inputs,
    device=self.device,
    dynamic=self.dynamic
)
shark_module.set_frontend("tflite-tosa")
# Case1: Use shark_importer default generate inputs
shark_module.compile()
mlir_results = shark_module.forward(inputs)
```

### DIFF
--- a/shark/shark_importer.py
+++ b/shark/shark_importer.py
@@ -1,48 +1,45 @@
 # Lint as: python3
 """SHARK Importer"""
 
-import iree.compiler.tflite as iree_tflite_compile
-import iree.runtime as iree_rt
 import numpy as np
 import os
-import sys
 import csv
-import tensorflow as tf
 import urllib.request
-from shark.shark_inference import SharkInference
-import iree.compiler.tflite as ireec_tflite
 from shark.iree_utils._common import IREE_TARGET_MAP
+import json
+from tank.model_utils_tflite import TFLiteModelUtil
 
 
 class SharkImporter:
     def __init__(
         self,
-        model_name: str = None,
-        model_path: str = None,
-        model_type: str = "tflite",
-        model_source_hub: str = "tfhub",
-        device: str = None,
-        dynamic: bool = False,
-        jit_trace: bool = False,
-        benchmark_mode: bool = False,
+        model_name,
+        model_type: str = "torch",
         input_details=None,
         output_details=None,
-        tank_url: str = None,
+        model_path=None,
     ):
         self.model_name = model_name
-        self.model_path = model_path
         self.model_type = model_type
-        self.model_source_hub = model_source_hub
-        self.device = device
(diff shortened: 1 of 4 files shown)


shark/shark_importer.py
```diff
@@ -51,60 +52,52 @@ def __init__(
...
+
+                inputs = self.generate_inputs(
+                    self.input_details
+                )  # device_inputs
+            self.setup_inputs(inputs)
+
+        elif self.model_type in ["tensorflow, tf, torch, pytorch"]:
+            print(self.model_type, " Not Implemented yet")
 
     def load_tflite_model(self):
```
pashu123, line 83: All the tf_lite related functions should go into tf_lite_utils.py. This will make this class clean. So you just have to call get_tf_lite_mlir(model_name, url ...) and it will get the model with all the metadata.

shark/shark_importer.py
```diff
@@ -117,26 +110,52 @@ def load_tflite_model(self):
...
+                input_type="tosa",
+                save_temp_iree_input=self.mlir_file,
+                target_backends=[IREE_TARGET_MAP["cpu"]],
+                import_only=False,
+            )
+        with open(self.mlir_file) as f:
+            self.mlir_model = f.read()
         return True
 
+    def setup_interpreter(self):
```
pashu123, line 139: All the tf-lite utils should go into tf_lite_utils.py.

pashu123 (changes requested):

pashu123:
Let's try to make the shark_importer clean and make sure that all the frontend-specific functions are put in a separate file/module.
This will reduce the overhead whenever a new frontend is introduced.

AmosLewis:
> Let's try to make the shark_importer clean and make sure that all the frontend-specific functions are put in a separate file/module. This will reduce the overhead whenever a new frontend is introduced.

ok

shark/shark_importer.py
AmosLewis: Done

shark/shark_importer.py
AmosLewis: Done

pashu123 (approved):
LGTM!

pashu123:
I will work on top of this. Thanks

PR #164, opened 2022-06-28
Rewrite&Move tflite examples to up tank dir

AmosLewis:
1. move tank/tflite tests up to tank
2. rewrite those tests to use new SharkImporter
3. disable tank/ dir tests on Mac


powderluv:
Why do we disable the tests for Mac ?

AmosLewis:
> Why do we disable the tests for Mac ?

E       ValueError: IREE compiler tool 'iree-import-tflite' is not installed (it should have been found in the python module 'iree.tools.tflite', typically installed via the package iree-tools-tflite).
[https://github.com/nod-ai/SHARK/runs/7123452064?check_suite_focus=true]

PR #169, opened 2022-06-30
Refactor TF tests for importer/runner split

monorimet:
Much more to do here but wanted to get some preliminary changes in ASAP.

todo:
-- fix repro paths for pytorch/TF tests
-- move tests out of frontend directories into tank/<modelname>/
-- benchmarks (API needs updating!)


monorimet:
Latest commit moves TF and PyTorch tests to model-specific directories in tank/. 
Model-specific directories are in camel case to avoid clashing with huggingface model paths.
Also changed all pytorch tests to use device argument instead of hardcoded "cpu", and changed "minilm_mlir" to "mlir_module".

Todo: benchmarks for tf and pytorch tests, save_temps for pytorch tests

monorimet:
PyTorch and TF tests are still in their own directories, this will be fixed in a follow-up commit.

pashu123 (approved):
Some comments. Rest LGTM!

PR #178, opened 2022-07-08
Update pytorch tests to support vulkan and cuda.

pashu123:
All the model validation pass except distilbert which is failing
in torch-mlir lowering. Also, added the mobilebert-uncased model to
the torch test suite.

### DIFF
--- a/shark/iree_utils/_common.py
+++ b/shark/iree_utils/_common.py
@@ -80,10 +80,8 @@ def check_device_drivers(device):
 # Installation info for the missing device drivers.
 def device_driver_info(device):
     if device in ["gpu", "cuda"]:
-        print(
-            "nvidia-smi not found, please install the required drivers from https://www.nvidia.in/Download/index.aspx?lang=en-in"
-        )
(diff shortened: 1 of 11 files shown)


shark/iree_utils/_common.py
```diff
@@ -80,10 +80,8 @@ def check_device_drivers(device):
 # Installation info for the missing device drivers.
 def device_driver_info(device):
     if device in ["gpu", "cuda"]:
-        print(
-            "nvidia-smi not found, please install the required drivers from https://www.nvidia.in/Download/index.aspx?lang=en-in"
-        )
+        return "nvidia-smi not found, please install the required drivers from https://www.nvidia.in/Download/index.aspx?lang=en-in"
```
powderluv, line 83: BTW I ran black without --check and it did the comments that way

powderluv (approved):

shark/iree_utils/_common.py
pashu123: Did you set `--line-length 127`?  because it would have failed in the CI

PR #186, opened 2022-07-16
Make PyTorch model tests importer-independent and add benchmarks.

monorimet:
Also adds torch.manual_seed(0) to the pytorch tests to make inputs consistent. This will have no effect if the downloader is used.

I would like to reimplement the importer section of these tests through a command-line option but let's get the basics down first.

Distilbert is skipped because we don't have a gs://shark_tank directory for it. I can help with getting TF tests equipped as well, but we will need the API to support TF. I'll keep working on it this weekend.

Any pytorch tests after this commit will have the option to write a .csv of benchmark results for comparing performances.

Usage:
```
pytest --benchmark tank/pytorch/*
```
Make sure to run tests from base SHARK/ directory.


powderluv (approved):
Awesome merge after checks are green

shark/shark_inference.py
```diff
@@ -60,22 +62,33 @@ def __init__(
...
-        # TODO: (Stanley) Update the shark_benchmark APIs.
-        self.shark_runner = SharkRunner(
-            self.mlir_module,
-            self.function_name,
-            self.device,
-            self.mlir_dialect,
-        )
+
+        if self.is_benchmark == True:
+            self.shark_runner = SharkBenchmarkRunner(
```
pashu123, line 78: Can we import the SharkBenchmarkRunner here?

PR #191, opened 2022-07-18
Disable tflite test on mac/darwin

AmosLewis:


powderluv:
Dont need for this disable right now.

PR #192, opened 2022-07-18
Import SharkBenchmarkRunner only if used.

monorimet:
Also reinstates a sys.exit(1) erroneously removed in shark_runner.py. This may solve issues with workflows.


monorimet:
Issue resolved. These features will be moved to another PR.

PR #199, opened 2022-07-22
Force TensorFlow to use CPU when generating model and golden values.

monorimet:
xfails will be updated once issues are filed for remaining failures.


pashu123:
@monorimet You can also add the tf part in shark_importer as well.

monorimet:
> @monorimet You can also add the tf part in shark_importer as well.

I put it in tank/masked_lm_tf.py so that it would work for benchmarks as well. Should it go in importer as well?

pashu123:
> > @monorimet You can also add the tf part in shark_importer as well.
> 
> I put it in tank/masked_lm_tf.py so that it would work for benchmarks as well. Should it go in importer as well?

Oh! It's fine over there.

AmosLewis (approved):
LGTM

PR #210, opened 2022-07-25
Remove op decomposition from the v_diffusion.py

vivekkhandelwal1:
The PyTorch decomposition for the op `aten.upsample_bilinear2d.vec`
is merged in the upstream repo and hence removed from this file.


pashu123 (approved):
LGTM!

PR #211, opened 2022-07-25
jupyter example of fine tuning with shark

dan-garvey:
to test, follow instructions on tank/README.md

May want to create a folder to put notebooks in instead of just floating at the top of the tank, lmk what you think.

### DIFF
--- a/tank/README.md
+++ b/tank/README.md
@@ -1,0 +1,13 @@
+To run the fine tuning example, from the root SHARK directory, run:
+
+```shell
+IMPORTER=1 ./setup_venv.sh
+source shark.venv/bin/activate
+pip install jupyter tf-models-nightly tf-datasets
+jupyter-notebook
+```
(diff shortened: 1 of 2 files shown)


powderluv:
can we merge this ?

tank/README.md
```diff
@@ -0,0 +1,13 @@
+To run the fine tuning example, from the root SHARK directory, run:
+
+```shell
+IMPORTER=1 ./setup_venv.sh
+source shark.venv/bin/activate
+pip install jupyter
```
powderluv, line 6: maybe add the `tf-models-nightly` here

powderluv (commented):
Can we move both files into tank/bert_fine_tuning/ ?

tank/README.md
dan-garvey: makes sense since this is the only thing that requires it. I'll add a try/catch on the import in case someone misses the instructions too with an error that provides a copy pastable pip install command

PR #214, opened 2022-07-27
[WIP] Add OPT to the tank

gpetters94:
This will require some refactoring of the model, but it should be able to use the same weights normal OPT.


requirements.txt
```diff
@@ -8,7 +8,7 @@ tqdm
 gsutil
 
 # generate_sharktank
-transformers==4.18.0
+transformers==4.19.4
```
powderluv, line 11: Actually I only did it for the importer. I didn't realize we added transformers to the base requirements.txt @monorimet

powderluv (approved):

PR #216, opened 2022-07-27
Add tf image classification auto model tests

AmosLewis:


generate_sharktank.py
```diff
@@ -4,15 +4,11 @@
...
-#         /albert_lite_base
-#         /...model_name...
-#       /tf
-#       /pytorch
+#       /albert_lite_base
+#       /...model_name...
 #
 
 import os
-import urllib.request
```
powderluv, line 15: would this cause the builders to hang ? All of them hung

AmosLewis: No, It wouldn't. this request is not used anymore, we use gsutil instead

powderluv (approved):

PR #217, opened 2022-07-28
Add hash of the mlir for checking upstream changes.

pashu123:
-- hash of the mlir files have been added in the shark_tank. (If there
is upstream changes, the model will be downloaded otherwise not).
-- The models have been placed in `~/.local/shark_tank/` folder as
opposed to the current folder not to redownload the files again.

### DIFF
--- a/generate_sharktank.py
+++ b/generate_sharktank.py
@@ -13,6 +13,8 @@ import csv
 import argparse
 from shark.shark_importer import SharkImporter
 import tensorflow as tf
+import hashlib
+import numpy as np
 
 visible_default = tf.config.list_physical_devices("GPU")
 try:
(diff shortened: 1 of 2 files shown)


pashu123:
@AmosLewis I have changed the model folder to `~/.local/shark_tank/` to make sure that every time these folders needn't be downloaded. I have also added hash for the PyTorch models. Could you add hash for tf and tf-lite models?

AmosLewis:
> @AmosLewis I have changed the model folder to `~/.local/shark_tank/` to make sure that every time these folders needn't be downloaded. I have also added hash for the PyTorch models. Could you add hash for tf and tf-lite models?

ok

monorimet (approved):
LGTM

PR #218, opened 2022-07-28
Fix GPU benchmarks on PyTorch model tests.

monorimet:
Run:
```
CUDA_VERSION=116 IMPORTER=1 ./setup_venv.sh
```
```
pytest --benchmark tank/pytorch/bert-base-uncased/bert-base-uncased_test.py
```
```
vim bench_results.csv
```


setup_venv.sh
```diff
@@ -108,6 +108,20 @@ fi
 
 $PYTHON -m pip install -e . --extra-index-url https://download.pytorch.org/whl/nightly/cpu -f https://github.com/llvm/torch-mlir/releases -f https://github.com/${RUNTIME}/releases
 
+if [[ ! -z "${CUDA_BENCHMARKS}" ]]; then
+  echo "${Yellow}Installing GPU benchmarking tools..."
+  if [[ $(uname -s) = 'Linux' ]]; then
+    echo "${Yellow}Linux detected.. installing Linux benchmarking tools"
+    $PYTHON -m pip uninstall -y torch torchvision
+    $PYTHON -m pip install --pre torch torchvision --extra-index-url https://download.pytorch.org/whl/nightly/cu116
```
powderluv, line 116: can we always just install the cuda version on linux and not key off CUDA_BENCHMARKS ?

monorimet: Yeah, I'll just add it under IMPORTER=1 which is required for TF benchmarks anyways.

monorimet: I removed the CUDA_BENCHMARKS switch. Now, if CUDA_VERSION is set, PyTorch+CUDA will be installed for CUDA_VERSION.

monorimet: Removed CUDA_VERSION.
torch+cuda116 will install for linux builds by default.

powderluv (approved):

PR #226, opened 2022-07-31
Update nightly.yml to remove IMPORTER=1 in package validation.

monorimet:


monorimet:
This should fix the issue with the nightly build. torch+cu116 is only installed if on Linux with IMPORTER=1.

powderluv (approved):

powderluv:
I'm not sure this is what we want since we may want to run pytest for GPU without IMPORTER=1. But let's unblock the build

monorimet:
Ok. The only thing we need from Pytorch+CUDA is the ability to run inference on the models with GPU (for benchmarks), right? In that case, it should be fine to guard it with IMPORTER=1, no?

PR #233, opened 2022-08-02
Check for MacStudio Target Value

PhaneeshB:
Need to check the vulkan target for MacStudio

### DIFF
--- a/.github/workflows/test-models.yml
+++ b/.github/workflows/test-models.yml
@@ -30,8 +30,6 @@ jobs:
             suite: cpu
           - os: MacStudio
             suite: gpu
-          - os: MacStudio
-            suite: vulkan
 
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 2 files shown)


pashu123:
`black --line-length=79 .`

powderluv (dismissed):
modulo linting

PhaneeshB:
linting error resolved

.github/workflows/test-models.yml
```diff
@@ -30,8 +30,6 @@ jobs:
             suite: cpu
           - os: MacStudio
             suite: gpu
-          - os: MacStudio
-            suite: vulkan
```
powderluv, line 34: i would switch this to cpu so we don't run on Mac CPU anylonger. But it can be a follow on commit

powderluv (approved):

PR #235, opened 2022-08-02
Mark XFail for M1 Vulkan Failures

PhaneeshB:
Xfail Marked for:
1. ConvBertModuleTest::test_module_static_vulkan
2. AlexnetModuleTest::test_module_static_vulkan
3. MobileBertModuleTest::test_module_static_vulkan
4. SqueezenetModuleTest::test_module_dynamic_vulkan


pashu123:
@PhaneeshB Can we create another test suite for mac vulkan? This will also xfail vulkan tests on other platforms.

powderluv (approved):
Let's land this to unblock. I think they just show up as xpass on Linux. Ideally we want to xfail on Darwin + Vulkan

PR #241, opened 2022-08-04
Upload benchmark results for every test-models workflow (excl. Vulkan)

monorimet:


monorimet:
try:
```
gsutil cp gs://iree-shared-files/nod-perf/bench_results/2022-08-04/ .
vim bench_results_gpu_da5b.csv
vim bench_results_cpu_da5b.csv
```

.github/workflows/test-models.yml
```diff
@@ -75,17 +80,21 @@ jobs:
       if: matrix.suite == 'cpu'
       run: |
         cd $GITHUB_WORKSPACE
-        PYTHON=python${{ matrix.python-version }} ./setup_venv.sh
+        PYTHON=python${{ matrix.python-version }} IMPORTER=1 ./setup_venv.sh
         source shark.venv/bin/activate
-        pytest -k 'cpu' --ignore=shark/tests/test_shark_importer.py --ignore=benchmarks/tests/test_hf_benchmark.py --ignore=benchmarks/tests/test_benchmark.py 
+        pytest --benchmark -k 'cpu' --ignore=shark/tests/test_shark_importer.py --ignore=benchmarks/tests/test_hf_benchmark.py --ignore=benchmarks/tests/test_benchmark.py 
+        gsutil cp ./bench_results.csv gs://iree-shared-files/nod-perf/bench_results/${DATE}/bench_results_cpu_${SHORT_SHA}.csv
+        rm bench_results.csv
```
powderluv, line 87: hopefully rm is not required :D

shark/shark_benchmark_runner.py
```diff
@@ -149,8 +149,8 @@ def benchmark_all_csv(
     ):
         self.setup_cl(inputs)
         field_names = [
-            "platform",
             "model",
+            "platform",
```
powderluv, line 153: some formatter requires the re-order ?

powderluv (approved):
minor nits

.github/workflows/test-models.yml
monorimet: Sending in a fix for this, as it shouldn't be there.

shark/shark_benchmark_runner.py
monorimet: Just changing the column order so that model comes first. This should do the trick

PR #251, opened 2022-08-09
Revert "Update test-models.yml"

PhaneeshB:
This reverts commit c2aa451767283166f68ea6045a223bff6f81a5a1.
TO keep a check on Mac Vulkan Tests


tank/resnet50_torch/resnet50_torch_test.py
```diff
@@ -88,6 +88,9 @@ def test_module_dynamic_gpu(self):
     @pytest.mark.skipif(
         check_device_drivers("vulkan"), reason=device_driver_info("vulkan")
     )
+    @pytest.mark.xfail(
```
pashu123, line 91: Instead of xfailing can we skip it if it's running on the mac-m1. I believe it runs on the other Vulkan backends.

pashu123:
You can get the device info from here: https://github.com/nod-ai/SHARK/blob/23619068eb829a7f9b709d47007db83bd5427609/shark/iree_utils/vulkan_utils.py#L20

tank/resnet50_torch/resnet50_torch_test.py
PhaneeshB: @pashu123 I've marked the tests xfail only for vulkan on macOS as suggested.

PhaneeshB:
Added changes to https://github.com/nod-ai/SHARK/pull/289

PR #253, opened 2022-08-10
Disable passing of sm_arch to iree-compile CL args by default.

monorimet:
This commit fixes numerics issues caused by TensorFloat32 execution on TensorFlow Masked LM models and PyTorch MobileBert.

TODO:
Re-enable the passing of this flag for when models require benchmarking -- this is a little sticky as we first validate, then use the same compiled .vmfb to benchmark on pytest runs.

Perhaps we can relax the tolerance when --benchmark is passed.

### DIFF
--- a/shark/parser.py
+++ b/shark/parser.py
@@ -47,16 +47,10 @@ parser.add_argument(
     default="./shark_tmp",
 )
 parser.add_argument(
-    "--save_mlir",
+    "--enable_tf32",
     default=False,
     action="store_true",
-    help="Saves input MLIR module to /tmp/ directory.",
-)
-parser.add_argument(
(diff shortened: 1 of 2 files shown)


shark/parser.py
```diff
@@ -47,16 +47,10 @@ def dir_file(path):
     default="./shark_tmp",
 )
 parser.add_argument(
-    "--save_mlir",
+    "--enable_tf32",
```
powderluv, line 50: adjusting tolerance based on this flag would be cool

powderluv (approved):

powderluv:
I assume we still set the target sm_arch right ? just that tf32 is disabled without the flag ?

PR #254, opened 2022-08-10
Update torch_mlir.compile API.

pashu123:
torch_mlir.compile API is updated and verified by compiling all the
torch models via generate_sharktank script.


dan-garvey (commented):
wow looks like the changes made it a lot easier

PR #257, opened 2022-08-10
[RFC] Parameterize devices for testing

raikonenfnu:
-Cleaner interface for device testing
-Easier to test new devices

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,3 +11,4 @@ gsutil
 pytest
 pytest-xdist
 Pillow
+parameterized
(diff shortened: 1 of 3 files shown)


raikonenfnu:
How the whole test looks like, if we specify do not specify any device:
Screen Shot 2022-08-10 at 1 22 05 PM

Still retains specifying device capability/ -k "<device_name>":
Screen Shot 2022-08-10 at 1 21 48 PM

powderluv:
So fancy 🎉 can static and dynamic also be a parameter?

monorimet:
This looks great. Chef's kiss 👨‍🍳🤌

raikonenfnu:
So added the parameterization of static/dynamic, although it did make it more complex since we need to do outerproduct to get all of the devices/staticity combinations. As well as a renaming function since adding more than one param messes up the function naming.  Also was thinking of ways to add xfail, but haven't thought of one that is elegant and simple yet. My idea was to feed the function a class or dictionary that we can query if it is marked as xfail and then add, something like:

```py
        if(xfail_cases.check(device, dynamic)):
             pytest.xfail(xfail[device])
```

PR #258, opened 2022-08-10
Make TF32 enabled by default and relax some tolerances if enabled.

monorimet:
`--enable_tf32` is now True by default and certain TF model tests have a switch for relaxing tolerances if TF32 calculations are enabled.


powderluv (approved):

monorimet:
Resolved by https://github.com/nod-ai/SHARK/pull/333

PR #259, opened 2022-08-11
split nightly pytest commands

dan-garvey:
prevents oom on the pytest command


monorimet (approved):
LGTM

powderluv:
I would do this: https://github.com/llvm/torch-mlir/blob/51bfe25c898e5f7a09294ccc012cfbf1327614c9/.github/workflows/oneshotSnapshotPackage.yml#L50-L58

So just call our "Validate Models on Shark Runtime" job

powderluv:
merging to unblock but lets do it to call out next

PR #264, opened 2022-08-13
Add benchmarks to MHLO miniLM and resnet50 and fix some issues with tolerances for TF32.

monorimet:
NOTE: benchmark reference numbers from TF on GPU are showing CPU results as GPUs are disabled in tf.config.
It's a bit tricky to navigate TF GPU execution/visibility so that will come in a follow-up.

### DIFF
--- a/shark/shark_benchmark_runner.py
+++ b/shark/shark_benchmark_runner.py
@@ -76,7 +76,7 @@ class SharkBenchmarkRunner(SharkRunner):
         torch_device = torch.device(
             "cuda:0" if self.device == "gpu" else "cpu"
         )
-        HFmodel, input, expected_output = get_torch_model(modelname)[:2]
+        HFmodel, input = get_torch_model(modelname)[:2]
         frontend_model = HFmodel.model
         frontend_model.to(torch_device)
         input.to(torch_device)
@@ -102,7 +102,9 @@ class SharkBenchmarkRunner(SharkRunner):
         import tensorflow as tf
         from tank.model_utils_tf import get_tf_model
 
-        model, input, expected_output = get_tf_model(modelname)
+        model, input, = get_tf_model(
+            modelname
+        )[:2]
         frontend_model = model
 
         for i in range(shark_args.num_warmup_iterations):
--- a/shark/shark_benchmark_runner.py
+++ b/shark/shark_benchmark_runner.py
@@ -19,7 +19,6 @@ from shark.iree_utils.benchmark_utils import (
     run_benchmark_module,
 )
 from shark.parser import shark_args
-from tank.model_utils import get_torch_model
 from datetime import datetime
 import time
 import csv
@@ -58,16 +57,17 @@ class SharkBenchmarkRunner(SharkRunner):
             input_tensors,
             mlir_dialect=self.mlir_dialect,
         )
-        print(self.benchmark_cl)
+        # print(self.benchmark_cl)
 
-    def benchmark_frontend(self, inputs, modelname):
+    def benchmark_frontend(self, modelname):
         if self.mlir_dialect in ["linalg", "torch"]:
             return self.benchmark_torch(modelname)
         elif self.mlir_dialect in ["mhlo", "tf"]:
-            return self.benchmark_tf(inputs, modelname)
+            return self.benchmark_tf(modelname)
 
     def benchmark_torch(self, modelname):
         import torch
+        from tank.model_utils import get_torch_model
 
         if self.device == "gpu":
             torch.set_default_tensor_type(torch.cuda.FloatTensor)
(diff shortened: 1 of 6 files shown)


tank/model_utils_tf.py
```diff
@@ -140,6 +210,8 @@ def get_causal_lm_model(hf_name, text="Hello, this is the default text."):
     #    gpus = tf.config.experimental.list_physical_devices("GPU")
     #    for gpu in gpus:
     #        tf.config.experimental.set_memory_growth(gpu, True)
+    BATCH_SIZE = 1
+    MAX_SEQUENCE_LENGTH = 16
```
powderluv, line 214: We want 128

shark/shark_benchmark_runner.py
```diff
@@ -76,7 +76,7 @@ def benchmark_torch(self, modelname):
         torch_device = torch.device(
             "cuda:0" if self.device == "gpu" else "cpu"
         )
-        HFmodel, input, act_out = get_torch_model(modelname)
+        HFmodel, input, _ = get_torch_model(modelname)
```
dan-garvey, line 79: If you know what this is, even if it doesn't get used, its better to give it a name, or not assign it and append "[:2]" to get_torch_model(modelname) (assuming it returns a list or tuple)

shark/shark_benchmark_runner.py
```diff
@@ -98,26 +98,32 @@ def benchmark_torch(self, modelname):
...
-        # return [
-        #    f"{shark_args.num_iterations/(end-begin)}",
-        #    f"{((end-begin)/shark_args.num_iterations)*1000}",
-        # ]
-        return ["n/a", "n/a"]
+    def benchmark_tf(self, modelname):
+        import tensorflow as tf
+        from tank.model_utils_tf import get_tf_model
+
+        model, input, _ = get_tf_model(modelname)
```
dan-garvey, line 105: same as above

tank/MiniLM-L12-H384-uncased_torch/MiniLM-L12-H384-uncased_torch_test.py
```diff
@@ -19,17 +20,7 @@ def create_and_check_module(self, dynamic, device):
...
-        # mlir_importer = SharkImporter(
-        #    model,
-        #    (input,),
-        #    frontend="torch",
-        # )
-        # minilm_mlir, func_name = mlir_importer.import_mlir(
-        #    is_dynamic=dynamic, tracing_required=True
-        # )
-
+        # shark_args.enable_tf32 = self.benchmark
```
dan-garvey, line 23: Delete or add a comment why it is commented and a TODO: with what needs to be done

tank/model_utils_tf.py
```diff
@@ -16,10 +16,48 @@
...
+    "mpnet-base",
+    "rembert",
+    "roberta-base",
+    "tapas-base",
+    "tiny-random-flaubert",
+    "xlm-roberta",
+]
+tfhf_models = [
+    "microsoft/MiniLM-L12-H384-uncased",
+]
```
dan-garvey, line 44: Maintaining this list seems irritating, is there a way it could be generated? Can be a TODO

tank/model_utils_tf.py
```diff
@@ -57,7 +95,7 @@ def get_TFhf_model(name):
         text,
         padding="max_length",
         truncation=True,
-        max_length=MAX_SEQUENCE_LENGTH,
+        max_length=128,
```
dan-garvey, line 98: Why is this changed when MAX_SEQUENCE_LENGTH still is 128? Ideally don't use numbers when you can use a named variable

dan-garvey (changes requested):
use git commit --amend instead of making 7 commits. Squash those so commit history is cleaner = )

tank/model_utils_tf.py
monorimet: yes -- maybe we can pull it from tank/tf/tf_model_list.csv
I have added a TODO.

tank/model_utils_tf.py
monorimet: This was a temporary fix to a problem that I have fixed in model_utils_tf. See new commit.

powderluv: You can still set the #define to 128. In general numbers in code should be #defines

powderluv (dismissed):

shark/shark_benchmark_runner.py
```diff
@@ -58,16 +57,17 @@ def setup_cl(self, input_tensors):
             input_tensors,
             mlir_dialect=self.mlir_dialect,
         )
-        print(self.benchmark_cl)
+        # print(self.benchmark_cl)
```
dan-garvey, line 60: dead

dan-garvey (approved):
+1

shark/shark_benchmark_runner.py
dan-garvey: (sorry if I missed this on the first pass)

PR #267, opened 2022-08-15
Improvements to pytest benchmarks.

monorimet:
Try:
```
IMPORTER=1 ONNX=1 ./setup_venv.sh
pytest --benchmark --onnx_bench tank/MiniLM-L12-H384-uncased
```

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -7,10 +7,10 @@ def pytest_addoption(parser):
         help="Pass option to benchmark and write results.csv",
     )
     parser.addoption(
-        "--onnx_bench"
+        "--onnx_bench",
         action="store_true",
         default="False",
-        help="Add ONNX benchmark results to pytest benchmarks."
+        help="Add ONNX benchmark results to pytest benchmarks.",
     )
     # The following options are deprecated and pending removal.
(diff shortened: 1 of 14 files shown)


monorimet:
Will squash to keep commit history clean.

conftest.py
```diff
@@ -1,22 +1,29 @@
...
     parser.addoption(
-        "--save_mlir",
+        "--benchmark",
         action="store_true",
         default="False",
-        help="Pass option to save input MLIR",
+        help="Pass option to benchmark and write results.csv",
     )
     parser.addoption(
-        "--save_vmfb",
```
powderluv, line 10: Don't want to leave this in ?

powderluv (approved):

conftest.py
monorimet: There's a .vmfb that gets saved locally without the arg here if I remember correctly.
I've been meaning to get rid of / rework the deprecated pytest options for a few weeks now. I'm going to go ahead and make sure it's tracked so we can get the option back in (or at least make it easier to tell where the .vmfb is being saved otherwise).

PR #286, opened 2022-08-22
Fix local artifact recognition and usage by SHARK downloader.

monorimet:


generate_sharktank.py
```diff
@@ -28,7 +29,8 @@
     pass
 
 # All generated models and metadata will be saved under this directory.
-WORKDIR = os.path.join(os.path.dirname(__file__), "gen_shark_tank")
+home = str(Path.home())
+WORKDIR = os.path.join(home, ".local/shark_tank/")
```
AmosLewis, line 33: line 242        
os.system(
            "gsutil cp -r ./gen_shark_tank/* gs://shark_tank/" + git_hash
        )

AmosLewis (changes requested):

generate_sharktank.py
monorimet: thanks!

shark/shark_downloader.py
```diff
@@ -110,7 +110,9 @@ def gs_download_model():
             np.load(os.path.join(model_dir, "upstream_hash.npy"))
         )
         if local_hash != upstream_hash:
-            gs_download_model()
```
dan-garvey, line 113: how does model downloading occur if not when the hash is nonmatching?

monorimet: If ~/.local/shark_tank/ is deleted, as directed by the print statement in check_dir_exists

AmosLewis (commented):
LGTM

AmosLewis (approved):

shark/shark_downloader.py
dan-garvey: can we add a flag that disables the hash auto download and leave it enabled by default?

PR #287, opened 2022-08-23
adds support to default to tuned model

dan-garvey:
currently setup for tf bert/resnet50 going to refactor test class to avoid having to add an argument to 50+ files


monorimet (dismissed):
Looks good! We can always pass device as tuned argument or will it eventually be specific target triples?

dan-garvey:
> Looks good! We can always pass device as tuned argument or will it eventually be specific target triples?

next patch will allow for target-triple

monorimet (approved):
Ok looks good.

PR #289, opened 2022-08-23
Simplified Testing Interface

PhaneeshB:
-Add support for parameterized test case generation.
-Add parameterized test case generation for each model.
-Re-Enable Mac Vulkan Test in CI

### DIFF
--- a/tank/albert-base-v2_tf/albert-base-v2_tf_test.py
+++ b/tank/albert-base-v2_tf/albert-base-v2_tf_test.py
@@ -1,6 +1,8 @@
 from shark.iree_utils._common import check_device_drivers, device_driver_info
 from shark.shark_inference import SharkInference
 from shark.shark_downloader import download_tf_model
+from tank.test_utils import get_valid_test_params, shark_test_name_func
+from parameterized import parameterized
 
 import iree.compiler as ireec
 import unittest
@@ -34,34 +36,10 @@ class AlbertBaseModuleTest(unittest.TestCase):
         self.module_tester = AlbertBaseModuleTester(self)
         self.module_tester.benchmark = pytestconfig.getoption("benchmark")
 
-    def test_module_static_cpu(self):
-        dynamic = False
-        device = "cpu"
-        self.module_tester.create_and_check_module(dynamic, device)
+    param_list = get_valid_test_params()
 
-    @pytest.mark.skipif(
-        check_device_drivers("gpu"), reason=device_driver_info("gpu")
-    )
-    def test_module_static_gpu(self):
-        dynamic = False
-        device = "gpu"
-        self.module_tester.create_and_check_module(dynamic, device)
-
-    @pytest.mark.skipif(
-        check_device_drivers("vulkan"), reason=device_driver_info("vulkan")
-    )
-    def test_module_static_vulkan(self):
-        dynamic = False
-        device = "vulkan"
-        self.module_tester.create_and_check_module(dynamic, device)
-
-    @pytest.mark.skipif(
-        check_device_drivers("intel-gpu"),
-        reason=device_driver_info("intel-gpu"),
-    )
-    def test_module_static_intel_gpu(self):
-        dynamic = False
-        device = "intel-gpu"
+    @parameterized.expand(param_list, name_func=shark_test_name_func)
+    def test_module(self, dynamic, device):
         self.module_tester.create_and_check_module(dynamic, device)
 
 
--- a/tank/alexnet_torch/alexnet_torch_test.py
+++ b/tank/alexnet_torch/alexnet_torch_test.py
@@ -1,8 +1,11 @@
 from shark.shark_inference import SharkInference
 from shark.iree_utils._common import check_device_drivers, device_driver_info
 from tank.model_utils import compare_tensors
+from shark.iree_utils.vulkan_utils import get_vulkan_triple_flag
 from shark.shark_downloader import download_torch_model
+from tank.test_utils import get_valid_test_params, shark_test_name_func
 
+from parameterized import parameterized
 import unittest
 import numpy as np
 import pytest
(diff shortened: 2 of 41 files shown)


tank/albert-base-v2_tf/albert-base-v2_tf_test.py
```diff
@@ -34,34 +36,16 @@ def configure(self, pytestconfig):
...
-    @pytest.mark.skipif(
-        check_device_drivers("intel-gpu"),
-        reason=device_driver_info("intel-gpu"),
-    )
-    def test_module_static_intel_gpu(self):
-        dynamic = False
-        device = "intel-gpu"
+    @parameterized.expand(param_list, name_func=shark_test_name_func)
+    def test_module(self, dynamic, device):
+        if device == "gpu":
```
raikonenfnu, line 43: you shouldn’t need to check for devices capability since it’s already part of get_valid_params

tank/alexnet_torch/alexnet_torch_test.py
```diff
@@ -57,49 +59,19 @@ def configure(self, pytestconfig):
...
+    @parameterized.expand(param_list, name_func=shark_test_name_func)
+    def test_module(self, dynamic, device):
+        if device == "gpu":
+            if check_device_drivers("gpu"):
+                pytest.skip(reason=device_driver_info("gpu"))
+        elif device == "vulkan":
+            if check_device_drivers("vulkan"):
+                pytest.skip(reason=device_driver_info("vulkan"))
+            else:
+                if dynamic == False:
```
raikonenfnu, line 73: although simple, is this the cleanest way we can use? I was thinking we can have some dict like data structure where it’s easy for developer to add or mark xfail cases, any thoughts @monorimet

PhaneeshB:
Before this change we added individual test case for each device type for every model which is difficult to maintain for a large number of models.

This change aims to create a simplified method to add tests for a new model.

Example bert-base-cased_torch

How test cases were added before:
```
    def test_module_static_cpu(self):
        dynamic = False
        device = "cpu"
        self.module_tester.create_and_check_module(dynamic, device)

    def test_module_dynamic_cpu(self):
        dynamic = True
        device = "cpu"
        self.module_tester.create_and_check_module(dynamic, device)

    @pytest.mark.skipif(
        check_device_drivers("gpu"), reason=device_driver_info("gpu")
    )
    def test_module_static_gpu(self):
        dynamic = False
        device = "gpu"
        self.module_tester.create_and_check_module(dynamic, device)

    @pytest.mark.skipif(
        check_device_drivers("gpu"), reason=device_driver_info("gpu")
    )
    def test_module_dynamic_gpu(self):
        dynamic = True
        device = "gpu"
        self.module_tester.create_and_check_module(dynamic, device)

    @pytest.mark.skipif(
        check_device_drivers("vulkan"), reason=device_driver_info("vulkan")
    )
    def test_module_static_vulkan(self):
        dynamic = False
        device = "vulkan"
        self.module_tester.create_and_check_module(dynamic, device)
```

How test cases will be added after:
```
    param_list = get_valid_test_params()

    @parameterized.expand(param_list, name_func=shark_test_name_func)
    def test_module(self, dynamic, device):
        if device == "gpu":
            if check_device_drivers("gpu"):
                pytest.skip(reason=device_driver_info("gpu"))
        elif device == "vulkan":
            if check_device_drivers("vulkan"):
                pytest.skip(reason=device_driver_info("vulkan"))
        self.module_tester.create_and_check_module(dynamic, device)
        
```

Here all the combinations of the test case args `(dynamic, device)` are populated by `get_valid_test_params()` and the test methods are auto-generated using this list.

To skip or xfail a particular test case, we just need to add a condition before the module tester is involving the args for the test

For example, to skip all `vulkan` tests for a model simply add the following 
```
if device == "vulkan":
     pytest.skip(reason="Your reason to skip the test")  
```

Similarly to xfail a test for `static` `gpu`

``` 
if device == "gpu" and dynamic == False:
        pytest.xfail(reason="Your reason to xfail the test")  
```

tank/alexnet_torch/alexnet_torch_test.py
dan-garvey: I'm adding support for this in a patch i'm working on

tank/albert-base-v2_tf/albert-base-v2_tf_test.py
PhaneeshB: Makes sense, I was following the code in test cases and added this check for capabilities which should be needed anymore.

tank/alexnet_torch/alexnet_torch_test.py
monorimet: Looks like Dan has some ideas about this. 
LGTM as long as we can xfail easily.

monorimet: It's also important to be able to easily designate a specific test case or a subset of test cases based on model, dynamic, device params.
We've been doing `pytest tank/albert-base-v2_torch/ -k "static_gpu"` , for example. As long as we can do that just as easily this looks great! :)

PhaneeshB: > It's also important to be able to easily designate a specific test case or a subset of test cases based on model, dynamic, device params. We've been doing `pytest tank/albert-base-v2_torch/ -k "static_gpu"` , for example. As long as we can do that just as easily this looks great! :)

This should not change as the final test cases will have same names as earlier, here we're eliminating the need to explicitly write all the test cases.

tank/alexnet_torch/alexnet_torch_test.py
```diff
@@ -57,49 +59,19 @@ def configure(self, pytestconfig):
...
-        device = "vulkan"
+    param_list = get_valid_test_params()
+
+    @parameterized.expand(param_list, name_func=shark_test_name_func)
+    def test_module(self, dynamic, device):
+        if device == "gpu":
+            if check_device_drivers("gpu"):
+                pytest.skip(reason=device_driver_info("gpu"))
+        elif device == "vulkan":
+            if check_device_drivers("vulkan"):
```
pashu123, line 70: Also, add a check for mac here. I see you have imported the target triple flag and xfail accordingly.

raikonenfnu:
FWIW, LGTM now :)

monorimet:
Yes, this looks great!

monorimet (dismissed):

tank/alexnet_torch/alexnet_torch_test.py
PhaneeshB: Will go ahead with this for now and make the required changes later on.

tank/alexnet_torch/alexnet_torch_test.py
PhaneeshB: > Also, add a check for mac here. I see you have imported the target triple flag and xfail accordingly.

Made changes to xfail vulkan tests specifically for Mac

monorimet (approved):
LGTM

PR #304, opened 2022-09-01
xfail vulkan unsupported op tests

PhaneeshB:
Support for some arith ops have been removed from llvm

-xfail all related test
-add Mac Vulkan test to CI


powderluv:
re-run after the latest nightly so you have all the changes.

PR #323, opened 2022-09-12
Update test-models.yml

monorimet:


monorimet:
This makes vulkan job use IMPORTER=1 as the torch install seems to fail without. I will get the non-importer setup working today.

powderluv (approved):
This will break metal I think. worth checking on the macstudio builder.

PR #333, opened 2022-09-15
creates abstract test case class

dan-garvey:


powderluv (approved):
This looks easier to scale

PR #355, opened 2022-09-22
update xfails for PyTorch DistilBERT

monorimet:
Fixes nightly failures while generating sharktank.


dan-garvey (approved):
May want to add xfail as a csv entry, and just xfail everything with that entry

monorimet:
> May want to add xfail as a csv entry, and just xfail everything with that entry

Yes, this would be cleaner. Our xfails need work for other reasons as well so I'll set them up as entries in a .csv instead. I'm thinking of doing this in `all_models.csv` for normal xfails(cpu/cuda/vulkan) and special ones (e.g. for specific vulkan triples) will probably still need to be written in this way.

PR #357, opened 2022-09-22
Fix issues with loading .vmfb into SharkInference

monorimet:


monorimet:
Some of this is @pashu123 's work -- we can sort out the commits correctly if desired

pashu123 (approved):
LGTM! Nice.

PR #360, opened 2022-09-23
Add sharkdynamo which combines shark with torchdynamo.

pashu123:


pashu123:
@powderluv Even for-loops are supported through torchdynamo. :smiley:

powderluv (approved):
Awesome. You may want to get rid of the old integraty Max had

powderluv (approved):

PR #361, opened 2022-09-23
Update model annotation tool

yzhang93:
Usage:
from shark.model_annotation import model_annotation
with create_context() as ctx:
  module = model_annotation(ctx, input_contents=..., config_path=..., search_op=...)

Example:
The example is to annotate the minilm model with GPU config file. 
python model_annotation.py /nodclouddata/vivian/minilm_model/model.mlir /nodclouddata/vivian/minilm_model/model_config.json

### DIFF
--- a/shark/model_annotation.py
+++ b/shark/model_annotation.py
@@ -12,22 +12,21 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import sys
 import json
 import os
-from typing import List, Dict
+import sys
+from typing import Dict, List
 
 from iree.compiler import ir
 from iree.compiler.transforms import ireec as ireec_trans
 
-MATMUL_OP_NAMES = set(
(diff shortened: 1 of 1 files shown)


powderluv (approved):
Please also add a readme

yzhang93:
> Please also add a readme

@powderluv Actually there's not much to write in a separate README file. It's just a function call to model_annotation(). I'll let @monorimet to decide where to store the tuned config files and where to call this function. I can provide the config files for tank models.

monorimet:
Looks good. I'll spend some time with this so I'm familiar enough to integrate.

1. You can `pip install black` and run `black --line-length 79 . ` from SHARK base directory for the formatting issues.

monorimet (dismissed):

yzhang93:
@monorimet Thanks! I've updated the format. Could you take a look again?

PR #375, opened 2022-09-28
Add suggestions to use ubuntu 20.04 or greater with SHARK runtime binary installations.

monorimet:


setup_venv.sh
```diff
@@ -63,7 +63,7 @@ if [[ $(uname -s) = 'Darwin' ]]; then
     torch_mlir_bin=true
   fi
 elif [[ $(uname -s) = 'Linux' ]]; then
-  echo "${Yellow}Linux detected"
+  echo "${Yellow}Linux detected. If you are using Ubuntu <20.04, binary installation of SHARK runtime is not supported -- consider upgrading."
```
powderluv, line 66: Maybe we can check lsb_release -d and print the warning if it is no Ubuntu 22.04

PR #377, opened 2022-09-28
Adds devices and shard_sizes attributes to model_annotation

IanNod:
- Adds function to provide for array attributes from lists
- Parses and adds annotations for devices and shard_sizes attributes for distributed models


sogartar:
I would prefer some tests for that.

PR #387, opened 2022-10-06
add support for bert training to generate_sharktank and pytest+benchmark

dan-garvey:
This has a lot of changes to the infra for generating tank and running benchmarks for training workload, so please review thoroughly.

As we add more support for training models, I imagine there can be some improvements to remove some boilerplate I've added here.


generate_sharktank.py
```diff
@@ -239,11 +265,11 @@ def is_valid_file(arg):
     if args.torch_model_csv:
         save_torch_model(args.torch_model_csv)
 
-    if args.tf_model_csv:
-        save_tf_model(args.tf_model_csv)
+    # if args.tf_model_csv:
```
monorimet, line 268: Is this commenting intentional? I thought these args were enabled by default and the script needed them.

tank/all_models.csv
```diff
@@ -1,34 +1 @@
...
-google/mobilebert-uncased,mhlo,tf,1e-2,1e-3,default
-google/vit-base-patch16-224,mhlo,tf,1e-2,1e-3,tf_vit
-hf-internal-testing/tiny-random-flaubert,mhlo,tf,1e-2,1e-3,default
-microsoft/MiniLM-L12-H384-uncased,mhlo,tf,1e-2,1e-3,tf_hf
-microsoft/layoutlm-base-uncased,mhlo,tf,1e-2,1e-3,default
-microsoft/mpnet-base,mhlo,tf,1e-2,1e-3,default
-albert-base-v2,linalg,torch,1e-2,1e-3,default
-alexnet,linalg,torch,1e-2,1e-3,default
-bert-base-cased,linalg,torch,1e-2,1e-3,default
 bert-base-uncased,linalg,torch,1e-2,1e-3,default
```
monorimet, line 1: Should we add a column here or a separate .csv to indicate which models should be tested on training?

dan-garvey: Oooh, didnt notice this, I was removing some to test something and I thought I reverted it before adding. Thanks

generate_sharktank.py
dan-garvey: sorry this is for testing only, good catch

PR #389, opened 2022-10-06
Move old test scripts out of base tank directory and add xfails.

monorimet:


powderluv (commented):
Is this moving everything inside tf/ pytorch/ etc ? We were trying to move all model outside of the frameworks so they just look at it as supported SHARK models -- without attribution to native framework.

monorimet:
> Is this moving everything inside tf/ pytorch/ etc ? We were trying to move all model outside of the frameworks so they just look at it as supported SHARK models -- without attribution to native framework.

The tests moved in this PR and the specific model directories are no longer conformant to our testing infrastructure. They are only being kept until they are integrated into the csv-based test suite. We should also remove or have a model blacklist for the currently unsupported models that have been excluded from the master list in `tank/all_models.csv`.
In general the tank directory can be cleaned up some, and a few .csvs can probably be consolidated but it would be about a day's work. 
I can do that cleanup in this patch if preferred, or leave it as it was.

powderluv (approved):
ok. eventually we don't want tank/tf tank/pytorch etc

dan-garvey:
> ok. eventually we don't want tank/tf tank/pytorch etc

eventually it will just be a csv file and some utils

PR #397, opened 2022-10-12
Update URL for IREE compiler/runtime install

monorimet:
Also: 
Fix assets path
xfail FlauBERT
Fix username in gh-pages release .yml

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -120,7 +120,7 @@ jobs:
         GITHUB_TOKEN: ${{ secrets.NODAI_INVOCATION_TOKEN }}
       with:
         release_id: ${{ steps.create_release.outputs.id }}
-        assets_path: ${GITHUB_WORKSPACE}/wheelhouse/nodai_*.whl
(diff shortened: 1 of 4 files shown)


.github/workflows/nightly.yml
```diff
@@ -120,7 +120,7 @@ jobs:
         GITHUB_TOKEN: ${{ secrets.NODAI_INVOCATION_TOKEN }}
       with:
         release_id: ${{ steps.create_release.outputs.id }}
-        assets_path: ${GITHUB_WORKSPACE}/wheelhouse/nodai_*.whl
```
dan-garvey, line 123: why?

dan-garvey (approved):
minor nit

.github/workflows/nightly.yml
monorimet: when we updated assets path last time it hasnt published with .whl so it may not be finding the wheels correctly. This may fix the path by removing the brackets

monorimet: it hasn't fixed it but will in follow-up patch as this has more urgent fixes

PR #402, opened 2022-10-13
Allow only one workflow job per branch

PhaneeshB:
Allow one job per branch and kill previous jobs to prevent jobs from queuing up.


monorimet (dismissed):
Nice. LGTM.

monorimet (approved):

PR #420, opened 2022-10-21
Add params for Stable Diffusion

vivekkhandelwal1:


qedawkins (approved):
Fix the linting issue with `black --line-length=79 shark/examples/shark_inference/stable_diffusion` and then it looks good

PR #423, opened 2022-10-23
Pass the flags to vae.

pashu123:


pashu123:
@flags for fp32 and fp16 added.

powderluv (approved):

PR #429, opened 2022-10-24
Dispatch benchmarks

Eliasj42:


README.md
```diff
@@ -121,6 +121,33 @@ pytest tank/test_models.py -k "MiniLM"
...
+```
+
+Output will include:
+- Inside the specified directory, there will be a directory for each dispatch (there will be mlir files for all dispatches, but only compiled binaries and benchmark data for the specified dispatches)
+- An .mlir file containing the dispatch benchmark 
+- A compiled .vmfb file containing the dispatch benchmark
+- An mlir file containing just the hal executable
+- A compiled .vmfb file of the hal executable
+- A .txt file containing benchmark output
+
```
yzhang93, line 149: Could you add another file to save the iree-benchmark command for each dispatch? Ideally, it's a .sh file that one can directly run and recreate the benchmark results.

yzhang93: It would be even better if you add a second .sh file and add --enable_rgp flag to iree-benchmark-module.

yzhang93 (approved):

PR #435, opened 2022-10-27
Serialize torch-mlir CAPI module as bytecode instead of string.

monorimet:


pashu123 (approved):
LGTM!

gpetters94 (dismissed):
LGTM

pashu123:
I think https://github.com/nod-ai/SHARK/blob/e9e61c789f367130d202ecf7ad272bc59729c5fe/shark/shark_downloader.py#L139 reading of files should also be changed. We will be uploading the bytecode file to tank and while downloading it should be reading the bytecode file rather than str files.

pashu123:
Also, all the string notation of mlir_module should be changed to actual module. https://github.com/nod-ai/SHARK/blob/01fdf5ee16ea91de91c7a503fb2cfc2ef56e6d4b/shark/shark_inference.py#L66

monorimet:
> Also, all the string notation of mlir_module should be changed to actual module.
> 
> https://github.com/nod-ai/SHARK/blob/01fdf5ee16ea91de91c7a503fb2cfc2ef56e6d4b/shark/shark_inference.py#L66

Thanks for the suggestions!!

monorimet:
@pashu123 had some issues reading in the bytecode from torch-mlir -- When reading it through SharkDownloader with `mode=rb` IREE-compile fails with `<stdin>:1:2: error: unexpected character` where the line referenced is something along the lines of `b'#map0 = affine_map<(d0, d1) -> (0, d1)>\n`

the TF and TFlite .mlir seem to work with either mode.

I'll keep working on this when I have a bit of downtime -- I think I need to make sure the code I referenced from Ramiro's change in IREE is implemented correctly for our use case (maybe writing the result of `bytecode_stream.getvalue()` to a file is different from passing it as an argument?)

pashu123 (dismissed):

pashu123 (approved):

PR #444, opened 2022-10-29
[WEB] Remove live preview and disable `resnet|albert_maskfill`

Shukla-Gaurav:
This commit removes live preview feature for now as it's not functional. This feature will be added in the next patch.


web/models/stable_diffusion/main.py
```diff
@@ -72,8 +72,8 @@ def get_models():
         IREE_EXTRA_ARGS += [
             "--iree-flow-enable-conv-nchw-to-nhwc-transform",
             "--iree-flow-enable-padding-linalg-ops",
-            "--iree-flow-linalg-ops-padding-size=16",
-            "--iree-flow-enable-iterator-space-fusion",
+            "--iree-flow-linalg-ops-padding-size=32",
+            "--iree-spirv-unify-aliased-resources=false",
```
qedawkins, line 76: The most recent update to shark has removed this flag, so we can drop it from here

web/models/stable_diffusion/main.py
```diff
@@ -250,16 +248,15 @@ def stable_diff_inf(
         if DEBUG:
             log_write.write(f"time={step_ms}ms")
         latents = scheduler.step(noise_pred, i, latents)["prev_sample"]
-        # scale and decode the image latents with vae
-        latents = 1 / 0.18215 * latents
```
qedawkins, line 254: For reference, looking at this shows why it was generating incorrect images with live preview. Because the latents are used as an input for the next step, doing this multiplication scaled up the values for the next input and likely overflowed the FP16 precision, as well as broke FP32 correctness :)

qedawkins (changes requested):
I only have one comment that needs fixing (removing the flag)

web/models/stable_diffusion/main.py
Shukla-Gaurav: yeah, got it. Thanks :)

Shukla-Gaurav:
@qedawkins all the comments have been addressed, please re-review. Thanks!

qedawkins (approved):
LGTM!

PR #461, opened 2022-11-02
Add Optional Support for RDC Profiling

JakopinA:


web/models/stable_diffusion/main.py
```diff
@@ -352,6 +366,9 @@ def stable_diff_inf(
     pil_images = [Image.fromarray(image) for image in images]
     out_img = pil_images[0]
 
+    if (args.enable_rtc_profiling):
```
pashu123, line 369: Did you test this? At least test running this file?

pashu123: The args are present in stable_args which is a different file.

web/models/stable_diffusion/main.py
```diff
@@ -14,6 +14,8 @@
 import time
 import os
 
+from shark.examples.shark_inference.stable_diffusion.main import start_profiling, end_profiling
```
pashu123, line 17: Please copy these functions here rather than calling these. Both are separate projects.

pashu123 (changes requested):

web/models/stable_diffusion/main.py
JakopinA: stable_args is one of the files I changed with this. I made some updates, would you mind checking them?

PR #482, opened 2022-11-13
Actually print the error message when SharkRunner can't initialize the driver

joker-eph:
Right now it would just terminate the process silently


powderluv (approved):

powderluv:
Thanks for PRs. Added you as a collaborator so you can merge once CIs pass.

PR #483, opened 2022-11-13
Fix dispatch saving code after 749a2c2d

joker-eph:
In 749a2c2d iree_device_map and iree_target_map have been made functions but not all of the uses were updated.


joker-eph:
@PhaneeshB

powderluv (approved):

PR #484, opened 2022-11-14
Enable XLA compiler for TF models

mariecwhite:
In order to enable the XLA compiler, functions need to be explicitly annotated. Latencies are lower with XLA enabled: https://www.tensorflow.org/xla


powderluv (approved):

powderluv:
We may want to fix up lint otherwise CI will fail

PR #492, opened 2022-11-15
Use correct TF device depending on configuration

mariecwhite:
TF visible devices was being hardcoded to use CPU. This PR fixes TF CPU/GPU paths so that GPU can be used when benchmarking CUDA. Each test much run in a separate process to get around TF global GPU memory management: https://github.com/tensorflow/tensorflow/issues/36465#issuecomment-582749350


powderluv:
@monorimet please review and merge

PR #496, opened 2022-11-18
Update stable diffusion models to point to new location.

pashu123:


pashu123:
@powderluv All the models have been updated with the date_tag i.e. 18nov.

PR #502, opened 2022-11-22
Add TF EfficientNet Model

mariecwhite:


dan-garvey (dismissed):
Looks good to me, I suppose in the future we might want to look at a better way to handle the input shapes of varying models instead of global variables but not really scope of this patch. Thanks for contributing!

monorimet (approved):
Looks great! Thanks.

PR #505, opened 2022-11-22
Update the sd models.

pashu123:


qedawkins (approved):
thanks for the quick fix!

PR #508, opened 2022-11-23
Add support for different schedulers.

pashu123:
Initial support for adding schedulers. This verifies the model running with the PNDM scheduler too.


powderluv (approved):
@dan-garvey  can we capture as a test too?

powderluv (approved):

PR #517, opened 2022-11-28
dump all isas with amdllpc

aldesilv:
cd ~/SHARK/shark/examples/shark_inference/stable_diffusion
python main.py --precision="fp16" --device="vulkan" --iree-vulkan-target-triple=rdna3-unknown-linux --no-load_vmfb --dispatch_benchmarks="all" --dispatch_benchmarks_dir="SD_dispatches" --dump_isa

### DIFF
--- a/shark/iree_utils/compile_utils.py
+++ b/shark/iree_utils/compile_utils.py
@@ -90,6 +90,17 @@ def create_dispatch_dirs(bench_dir, device):
                 )
 
 
+def dump_isas(bench_dir):
+    for d_ in os.listdir(bench_dir):
+        if os.path.isdir(f"{bench_dir}/{d_}"):
+            for f_ in os.listdir(f"{bench_dir}/{d_}"):
+                if f_.endswith(".spv"):
+                    os.system(
+                        f"amdllpc -gfxip 11.0 {bench_dir}/{d_}/{f_} -v > \
+                         {bench_dir}/{d_}/isa.txt"
+                    )
+
+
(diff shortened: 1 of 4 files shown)


shark/iree_utils/compile_utils.py
```diff
@@ -90,6 +90,17 @@ def create_dispatch_dirs(bench_dir, device):
                 )
 
 
+def dump_isas(bench_dir):
+  for d_ in os.listdir(bench_dir):
+      if os.path.isdir(f"{bench_dir}/{d_}"):
+          for f_ in os.listdir(f"{bench_dir}/{d_}"):
+              if f_.endswith('.spv'):
+                  os.system(
+                      f"/usr/bin/amdllpc -gfxip 11.0 {bench_dir}/{d_}/{f_} -v > \
```
powderluv, line 99: can we just call `amdllpc` so it can be anywhere in your path not just /usr/bin ?

powderluv:
Please add an example usage in the commit message. and fix up lint

shark/iree_utils/compile_utils.py
aldesilv: amdllpc: command not found, even though /usr/bin is in $PATH. also I have to use sudo /usr/bin/amdllpc ... or else Permission denied

powderluv: Wait I may need to set perms on amdllpc. You shouldn't have to use sudo

powderluv (approved):

PR #551, opened 2022-12-04
add Rust compiler to reqs for windows

cpietsch:


powderluv:
Btw why do we need rust ?

cpietsch:
On windows I got an error while installing transformers, asking for the rust compiler.
But that could also be an error on my side because I used python 3.11 an maybe it was trying to build from source.

powderluv:
Yeah we are still flushing out 3.11 support.  Will test this out once we get 3.11 working.  

Thanks for the PR.

PR #553, opened 2022-12-05
[WEB] Add command line args to shark web

Shukla-Gaurav:
1. Now the server can be launched with command line args.
2. The `precision` and `scheduler` parameters are now part of command line args instead of UI.
3. Add vae encode model wrapper.

Signed-Off-by: Gaurav Shukla


qedawkins:
Before reviewing, why are we moving the precision and scheduler parameters out of the UI? Is there a plan to move it back in later?

Shukla-Gaurav:
> Before reviewing, why are we moving the precision and scheduler parameters out of the UI? Is there a plan to move it back in later?

As discussed, we are trying to keep only those UI elements which are more relevant to artistic point of view.

powderluv (approved):

powderluv:
it also prevents downloading and compiling fp32 if you only want to use fp16

PR #567, opened 2022-12-06
Enable f32 path for SD

PhaneeshB:
Changes to download model from HF acc to precision
Takes half precision cuda path for fp16 and defaults to fp32 
No support for other precisions (int8) as of now.


pashu123 (approved):
LGTM.

PR #588, opened 2022-12-08
Disable SHARK-Runtime flags if USE_IREE=1 specified during setup.

monorimet:


setup_venv.sh
```diff
@@ -96,6 +96,7 @@ fi
 if [[ -z "${USE_IREE}" ]]; then
   RUNTIME="https://nod-ai.github.io/SHARK-Runtime/pip-release-links.html"
```
powderluv, line 97: You probably want to add a rm .use-iree here to handle the case of switching between one and another

powderluv (changes requested):

powderluv (approved):
Keep an eye out to see if the rm fails if no file exists

monorimet:
> Keep an eye out to see if the rm fails if no file exists

Looks like it just complains that the file isnt found but no hard fails

PR #589, opened 2022-12-08
 Autodetect cpu count for runtime flags.

monorimet:


powderluv (commented):
I think this merged the previous commit. CPU changes look good

shark/iree_utils/benchmark_utils.py
```diff
@@ -73,6 +74,10 @@ def build_benchmark_args(
     mlir_input_types = tensor_to_type_str(input_tensors, mlir_dialect)
     for mlir_input in mlir_input_types:
         benchmark_cl.append(f"--function_input={mlir_input}")
+    if device == "cpu":
+        num_cpus = get_cpu_count()
+        if num_cpus is not None:
+            benchmark_cl.append(f"--task_topology_max_group_count={num_cpus}")
```
powderluv, line 80: we should probably cap this at a max of 32

PR #602, opened 2022-12-10
Move most xfails to entries in tank/all_models.csv

monorimet:
This accomplishes a few improvements and optimizations to the process of marking test cases as expected to fail. It also prevents the need to run formatting checks, so unexpected regressions can easily be accounted for.


powderluv (dismissed):

.github/workflows/test-models.yml
```diff
@@ -123,7 +123,7 @@ jobs:
         echo "VULKAN SDK PATH with setup: $VULKAN_SDK"
         echo $PATH
         pip list | grep -E "torch|iree"
-        pytest --ci --ci_sha=${SHORT_SHA} --local_tank_cache="/Volumes/builder/anush/shark_cache" tank/test_models.py -k vulkan --update_tank
+        pytest -s --ci --ci_sha=${SHORT_SHA} --local_tank_cache="/Volumes/builder/anush/shark_cache" tank/test_models.py -k vulkan --update_tank
```
powderluv, line 126: this path `/Volumes/builder/anush/shark_cache` is only on the mac builder

monorimet: Right-- the other builders are pointing to `/data/anush/shark_cache/` . Is this incorrect? I thought I had these configured correctly.

monorimet: follow-up - weren't these all supposed to be pointing towards an external volume located at `data/anush/` for linux and `/Volumes/builder/anush/` for the mac builder?
Let me know if this isn't the case so I can fix it.

Issue #601, opened 2022-12-10
[web] loading timer hard-coded to 50 steps


Ian2400:
The loading timer will act as if it is running to 50 steps regardless of the number of steps chosen in the web UI

PR #605, opened 2022-12-10
[WEB] Iterative Output for Demo

cpietsch:
Wouldn't it be cool to see the progress image while running diffusion ?

Added support for [Gradio's Iterative Output](https://gradio.app/key_features/#iterative-outputs) to the web demo.
This might be even nicer than having a progressbar #601, though we could also add a seperate progressbar mis-using a [label](https://gradio.app/docs/#label) or waiting for [tqdm integration](https://github.com/gradio-app/gradio/issues/340).

Update currently set to every 20 step ``iterative_output_step = 20``

gradio

### DIFF
--- a/web/index.py
+++ b/web/index.py
@@ -166,6 +166,7 @@ with gr.Blocks(css=demo_css) as shark_web:
             outputs=[generated_img, std_output],
         )
 
+shark_web.queue()
 shark_web.launch(
     share=False,
     server_name="0.0.0.0",
--- a/web/index.py
+++ b/web/index.py
@@ -33,17 +33,19 @@ demo_css = """
 .gradio-container {background-color: black}
 .container {background-color: black !important; padding-top:20px !important; }
 #ui_title {padding: 10px !important; }
-#top_logo {background-color: transparent; border-radius: 0 !important; border: 0; } 
-#demo_title {background-color: black; border-radius: 0 !important; border: 0; padding-top: 50px; padding-bottom: 0px; width: 460px !important;} 
-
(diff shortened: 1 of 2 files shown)


powderluv:
Thank you for the contribution. We used to have the "live preview" option before and we dropped it. If there is interest in having it we can add it and merge this change. But I think we should have it as a radio button turned off by default.

cpietsch:
will add the radio button

cpietsch:
added a dropdown
image

powderluv:
Do you forsee any other option in the drop down ? If not best to have it as a radio button. It should be there in prior git history and the `rdna` branch probably has it

powderluv:
btw if you view it in dark mode the UI looks a little better (or set your windows theme to dark). We need to fix that.

cpietsch:
I added the dropdown, because you can choose the steps instead of a fixed value
``choices=["None", "5", "10", "20"],``
https://github.com/cpietsch/SHARK/blob/patch-2/web/index.py#L115

powderluv:
Thanks for the PR. Just a heads up we plan to move the UI into the main stable Diffusion script so expect some churn but then you can interchangeably just cli and webui.

powderluv:
Please fix the CI lint failures. Just need to run black

cpietsch:
fixed lint failures
great to see moving the web stuff into the main script - first I was confused seeing 2 times the diffusion script in the repo

powderluv:
finally the merge of the scripts happened. Lets close this and open a a new one if required.

PR #610, opened 2022-12-11
Initialize TF models locally

mariecwhite:
Initializing models globally causes TF to hang when XLA is enabled.


monorimet (approved):
LGTM!

PR #624, opened 2022-12-14
add flag for toggling vulkan validation layers

PhaneeshB:
By default the flag is enable only in Debug builds
This PR defaults the flag to false for Stable Diffusion Inference. 

Also, categorically rearranges args used for SD.


powderluv (dismissed):

powderluv (approved):
oh i thought this had already landed.

PR #628, opened 2022-12-14
pythonize uname for cpu target triple in windows

PhaneeshB:
Uses python uname utility for target triple manufacturing for cpu.

Draft because: Needs correct variation of target triple for windows supported by IREE


powderluv (approved):
Windows triple seems right

PR #634, opened 2022-12-14
Update tuned model SD v2.1base

yzhang93:


shark/examples/shark_inference/stable_diffusion/opt_params.py
```diff
@@ -23,7 +23,8 @@ def get_unet():
     if args.precision == "fp16":
         if args.use_tuned:
             bucket = "gs://shark_tank/vivian"
-            model_name = "unet_1dec_fp16_tuned"
+            if args.version == "v2.1base":
+                model_name = "unet2base_8dec_fp16_tuned"
```
qedawkins, line 27: this will cause --use_tuned to break if version v2.1base is not specified because model_name won't be set to anything.

shark/examples/shark_inference/stable_diffusion/opt_params.py
```diff
@@ -23,7 +23,8 @@ def get_unet():
     if args.precision == "fp16":
         if args.use_tuned:
             bucket = "gs://shark_tank/vivian"
-            model_name = "unet_1dec_fp16_tuned"
+            if args.version == "v2.1base":
```
pashu123, line 26: We can leave the v1.4 version as it is. `if args.version == "v1.4": model_name = "un.....1dec..."`.

qedawkins (changes requested):

shark/examples/shark_inference/stable_diffusion/opt_params.py
yzhang93: @qedawkins Isn't the default args.version already set as "v2.1base"? It won't cause break I think?

shark/examples/shark_inference/stable_diffusion/opt_params.py
yzhang93: Done

shark/examples/shark_inference/stable_diffusion/opt_params.py
qedawkins: as in there's no fallthrough. So if args.version != v2.1base or v1.4 then model_name won't be set to anything and the python script will crash. It's fine now though because v1.4 is set properly.

qedawkins (approved):

PR #651, opened 2022-12-16
[SD][WEB] Deduce vulkan-target-triple in the presence of multiple cards

Shukla-Gaurav:
1. Get the correct vulkan-target-triple for a specified device in the presence of multiple cards.
2. Use tuned unet model for rdna3 cards.


PhaneeshB (approved):
LGTM

PR #664, opened 2022-12-19
[SD][CL] Disable print at every iteration.

raikonenfnu:
Printing might incur extra time to runtime. Hence, we add a flag to hide it. To disable printing please set this flag `--hide_steps`.


pashu123 (approved):

powderluv (commented):
Let's toggle the default

PR #667, opened 2022-12-19
Add max prompt len 64 f16 untuned SD 

PhaneeshB:
Model added to support max prompt len of 64.
precision fp16.
Untuned.


pashu123 (approved):
LGTM!

PR #687, opened 2022-12-20
[SD] Add a json file for model names information.

Shukla-Gaurav:
This commit simplifies the code to identify the model name for a particular set of flags. This is achieved by introducing a json file that stores the model names information. The models are uploaded in gcloud with these names.


Shukla-Gaurav:
This simplifies only the web part. Once https://github.com/nod-ai/SHARK/pull/685/ will land, I will update the CLI part as well.
@powderluv @PhaneeshB

web/models/stable_diffusion/opt_params.py
```diff
@@ -25,156 +26,94 @@
...
+        f"SD/{args.version}/vae/{args.precision}/length_77{is_tuned}{is_base}"
+    )
+    model_name, iree_flags = get_params(model_key)
+    if args.use_tuned:
+        bucket = "gs://shark_tank/vivian"
         iree_flags += [
             "--iree-flow-enable-padding-linalg-ops",
             "--iree-flow-linalg-ops-padding-size=32",
+            "--iree-flow-enable-conv-img2col-transform",
+            "--iree-flow-enable-conv-winograd-transform",
```
qedawkins, line 89: @yzhang93 can correct me if I'm wrong but I don't think we should be doing winograd like this.

yzhang93: @qedawkins The current VAE tuned model is just with winograd attributes. So we'll need to use these flags.

web/models/stable_diffusion/cache_objects.py
```diff
@@ -18,8 +18,8 @@
 
 model_config = {
     "v2": "stabilityai/stable-diffusion-2",
-    "v2.1base": "stabilityai/stable-diffusion-2-1-base",
-    "v1.4": "CompVis/stable-diffusion-v1-4",
+    "v2_1base": "stabilityai/stable-diffusion-2-1-base",
```
yzhang93, line 21: Why not use -v2 and -v2-1-base? It's better to keep the same format for all names.

web/models/stable_diffusion/opt_params.py
qedawkins: It should do winograd regardless of the flag if it's annotated though, so this is forcing any non-winograd convs to go down winograd (although since it's tuned there might not be any convs left, so this flag doesn't matter?)

web/models/stable_diffusion/opt_params.py
```diff
@@ -25,156 +26,94 @@
...
+
+
+def get_unet():
+    # Tuned model is present only for `fp16` precision.
+    is_tuned = "/tuned" if args.use_tuned else "/untuned"
+    model_key = f"SD/{args.version}/unet/{args.precision}/length_{args.max_length}{is_tuned}"
+    model_name, iree_flags = get_params(model_key)
+    if args.use_tuned:
+        bucket = "gs://shark_tank/vivian"
+        return get_shark_model(bucket, model_name, iree_flags)
```
yzhang93, line 55: The tuned model is under condition `precision == "fp16"`.  We should add some checks of the precision.

Shukla-Gaurav: The models.json does not contain tuned fp32 so it will report that the model is not present.
What we can do is: make args.use_tuned=False explicitly if precision is fp32 for now in order to avoid any termination.

web/models/stable_diffusion/opt_params.py
yzhang93: @qedawkins I'm not sure I understand. Currently this model is the original mlir + some convs marked `iree_winograd_conv`. The mlir wasn't dump after certain pass, and doesn't have lowering configs. So I think these flags are needed for lowering?

web/models/stable_diffusion/cache_objects.py
Shukla-Gaurav: I have used _ as separator everywhere to keep same format. But yes, we can decide on formats having more readability, as they may get exposed to UI as well someday.

web/models/stable_diffusion/opt_params.py
yzhang93: Sounds good!

web/models/stable_diffusion/opt_params.py
Shukla-Gaurav: @yzhang93 @qedawkins So should we remove these flags from tuned, as for tuned the optimizations should have been done already?

powderluv (approved):

web/models/stable_diffusion/opt_params.py
Shukla-Gaurav: Keeping the flags as it is.

PR #692, opened 2022-12-21
Modify model annotation tool to walk through ops by shape

yzhang93:


shark/model_annotation.py
```diff
@@ -89,37 +121,167 @@ def walk_children(
...
+                int(kw),
+                int(f),
+                int(oh),
+                int(ow),
+                int(dilation),
+                int(stride),
+                int(pad),
+            ]
+
+        elif op.name in ["linalg.conv_2d_nhwc_hwcf"]:
```
qedawkins, line 239: I'm assuming we've never tuned nchw_fchw and that's why it's not present. Correct me if it's missing unintentionally though.

qedawkins (approved):

shark/model_annotation.py
yzhang93: Yes, you are right. I don't think we need to directly tune nchw_fchw ops.

PR #707, opened 2022-12-21
[SD] Update all the utilities to make web and CLI codebase closer

Shukla-Gaurav:
At this point, all the utilities of SD web and CLI are exactly same. 
Also the `web/models/stable_diffusion/main.py` can be directly executed.

### DIFF
--- a/web/models/stable_diffusion/cache_objects.py
+++ b/web/models/stable_diffusion/cache_objects.py
@@ -7,17 +7,17 @@ from diffusers import (
     EulerDiscreteScheduler,
 )
 from models.stable_diffusion.opt_params import get_unet, get_vae, get_clip
-from models.stable_diffusion.utils import set_iree_runtime_flags
+from models.stable_diffusion.utils import set_init_device_flags
 from models.stable_diffusion.stable_args import args
 from models.stable_diffusion.schedulers import (
     SharkEulerDiscreteScheduler,
 )
 
-# set iree-runtime flags
-set_iree_runtime_flags()
+# initial settings.
+set_init_device_flags()
 
 model_config = {
-    "v2": "stabilityai/stable-diffusion-2",
+    "v2_1": "stabilityai/stable-diffusion-2-1",
     "v2_1base": "stabilityai/stable-diffusion-2-1-base",
     "v1_4": "CompVis/stable-diffusion-v1-4",
 }
@@ -49,6 +49,7 @@ schedulers["SharkEulerDiscrete"] = SharkEulerDiscreteScheduler.from_pretrained(
 )
 schedulers["SharkEulerDiscrete"].compile()
 
+
 cache_obj = dict()
 # cache vae, unet and clip.
 (
@@ -61,7 +62,7 @@ cache_obj = dict()
 cache_obj["tokenizer"] = CLIPTokenizer.from_pretrained(
     "openai/clip-vit-large-patch14"
 )
-if args.version == "v2_1base":
+if args.version != "v1_4":
     cache_obj["tokenizer"] = CLIPTokenizer.from_pretrained(
-        "stabilityai/stable-diffusion-2-1-base", subfolder="tokenizer"
+        model_config[args.version], subfolder="tokenizer"
     )
--- a/web/models/stable_diffusion/main.py
+++ b/web/models/stable_diffusion/main.py
@@ -6,10 +6,29 @@ from models.stable_diffusion.cache_objects import (
     cache_obj,
     schedulers,
 )
+from models.stable_diffusion.utils import set_init_device_flags
 from models.stable_diffusion.stable_args import args
 from random import randint
 import numpy as np
 import time
+import sys
+
+
+# Helper function to profile the vulkan device.
+def start_profiling(file_path="foo.rdc", profiling_mode="queue"):
+    if args.vulkan_debug_utils and "vulkan" in args.device:
+        import iree
+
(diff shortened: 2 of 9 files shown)


Shukla-Gaurav:
@qedawkins Can you please help me test this patch? 
This will help a lot to merge the CLI and web.

Shukla-Gaurav:
@powderluv Once tested without any perf regression, the web folder will contain the cli as well.

Shukla-Gaurav:
Will test it tomorrow thoroughly before merging.

web/models/stable_diffusion/cache_objects.py
```diff
@@ -43,11 +47,20 @@
...
+if args.version == "v2_1base":
+    schedulers[
+        "SharkEulerDiscrete"
+    ] = SharkEulerDiscreteScheduler.from_pretrained(
+        model_config[args.version],
+        subfolder="scheduler",
+    )
+    schedulers["SharkEulerDiscrete"].compile()
+else:
+    schedulers["SharkEulerDiscrete"] = EulerDiscreteScheduler.from_pretrained(
```
qedawkins, line 59: I don't see why the SharkEulerDiscrete scheduler is exclusive to v2.1base.

web/models/stable_diffusion/main.py
```diff
@@ -10,6 +10,24 @@
 from random import randint
 import numpy as np
 import time
+import sys
+
+
+# Helper function to profile the vulkan device.
+def start_profiling(file_path="foo.rdc", profiling_mode="queue"):
```
qedawkins, line 17: We should check whether anyone is using this feature and if not, drop it. Another option would be to make it available for all models via SharkRunner.

web/models/stable_diffusion/utils.py
```diff
@@ -168,23 +167,27 @@ def map_device_to_name_path(device, key_combination=3):
...
+            f"Found device {device_name}. Using target triple {args.iree_vulkan_target_triple}."
+        )
+
+        # use tuned models only in the case of stablediffusion/fp16 and rdna3 cards.
+        if (
+            args.variant != "stablediffusion"
+            or args.precision != "fp16"
+            or "rdna3" not in args.iree_vulkan_target_triple
+        ):
             args.use_tuned = False
```
qedawkins, line 187: nit: add a warning/log similar to the one added below if use_tuned was set to true.

web/models/stable_diffusion/utils.py
```diff
@@ -168,23 +167,27 @@ def map_device_to_name_path(device, key_combination=3):
...
 
         if args.use_tuned:
             print("Using tuned models for rdna3 card")
     else:
-        if args.use_tuned:
-            print("Tuned models not currently supported for device")
-            args.use_tuned = False
+        args.use_tuned = False
+        print("Tuned models are currently not supported for this setting.")
```
qedawkins, line 193: I think the message here is still only necessary if use_tuned is true.

qedawkins (changes requested):
Still doing some e2e verifications, but I have a few comments first.

web/models/stable_diffusion/main.py
```diff
@@ -162,3 +184,25 @@ def stable_diff_inf(
     text_output += f"\nTotal image generation time: {total_time:.4f}sec"
 
     return pil_images[0], text_output
+
+
+if __name__ == "__main__":
```
qedawkins, line 189: This won't work due to two different entry points with different relative imports (index.py and models/stable_diffusion/main.py)
```
Traceback (most recent call last):
  File "/home/quinn/nod/SHARK/web/models/stable_diffusion/main.py", line 5, in <module>
    from models.stable_diffusion.cache_objects import (
ModuleNotFoundError: No module named 'models'
```
This is why we need #694 first to enable absolute imports for all of the relevant SD code.

qedawkins (changes requested):

web/models/stable_diffusion/main.py
Shukla-Gaurav: @qedawkins Let's land #694 first then. Thanks!

web/models/stable_diffusion/cache_objects.py
Shukla-Gaurav: Fixed. Thanks!

web/models/stable_diffusion/utils.py
Shukla-Gaurav: Added same message.

web/models/stable_diffusion/main.py
Shukla-Gaurav: Ok, so I have removed this second entry point from this PR as it is already there in #694. Now this PR is good to go before #694. I will take a look at the merging stuff with #694.

web/models/stable_diffusion/main.py
Shukla-Gaurav: yeah, we can discuss on this and put it at the right place.

Shukla-Gaurav:
@PhaneeshB @qedawkins I have tested it all and this is good to go. After this I will look into the movement PR #694.

PhaneeshB (approved):
LGTM

powderluv (approved):
Please resolve all comments. Please also generate an exe and validate the models work with the exe and then land

Shukla-Gaurav:
> Please resolve all comments. Please also generate an exe and validate the models work with the exe and then land

all comments are addressed and .exe is working fine.

PR #708, opened 2022-12-22
[SharkInference] Make SharkInference compile the entire module

Abhishek-Varma:
-- Prviously SharkInference was compiling and providing run APIs
   for a harcoded function with function name "forward".
-- This commit makes the compiling functionality generic and now
   any function being defined within the module can be run.
-- It also creates an API to fetch all the function names defined
   within the compiled module.

### DIFF
--- a/shark/iree_utils/compile_utils.py
+++ b/shark/iree_utils/compile_utils.py
@@ -234,7 +234,6 @@ def compile_module_to_flatbuffer(
     module,
     device,
     frontend,
-    func_name,
     model_config_path,
     extra_args,
     model_name="None",
@@ -277,7 +276,7 @@ def compile_module_to_flatbuffer(
     return flatbuffer_blob
 
 
-def get_iree_module(flatbuffer_blob, device, func_name):
+def get_iree_module(flatbuffer_blob, device):
     # Returns the compiled module and the configs.
     config = get_iree_runtime_config(device)
     vm_module = ireert.VmModule.from_flatbuffer(
@@ -285,7 +284,7 @@ def get_iree_module(flatbuffer_blob, device, func_name):
     )
     ctx = ireert.SystemContext(config=config)
     ctx.add_vm_module(vm_module)
-    ModuleCompiled = ctx.modules.module[func_name]
+    ModuleCompiled = ctx.modules.module
     return ModuleCompiled, config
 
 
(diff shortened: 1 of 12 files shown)


Abhishek-Varma:
Hi @powderluv @pashu123 - I've raised this draft PR to help move the checkpoint update task further.

I'm also looking at :-
1. Other places where it might break the pieces.
2. Integrating and testing it on yesterday's CLIP model.
3. Also work on having an `iree-opt ----iree-import-ml-program` as an API.
4. Cleanup/Refine further.

Will take some time here.

shark/shark_inference.py
```diff
@@ -138,21 +133,25 @@ def compile(self, extra_args=[]):
             os.system(f"rm -rf {self.temp_dispatch_benchmarks_dir}")
 
     # inputs are considered to be tuple of np.array.
-    def forward(self, inputs: tuple, send_to_host=True):
-        return self.shark_runner.run(inputs, send_to_host)
+    def __call__(self, function_name, inputs: tuple, send_to_host=True):
```
pashu123, line 136: The `function_name` can be annotated as a `str`.

shark/shark_runner.py
```diff
@@ -87,15 +83,19 @@ def __init__(
...
             self.iree_compilation_module,
+            function_name,
             inputs,
             self.iree_config,
             self.mlir_dialect,
             send_to_host,
         )
+    
+    # Get all function names defined within the compiled module.
+    def get_functions_in_module(self):
```
pashu123, line 100: Nice addition.

pashu123:
@Abhishek-Varma The draft PR really looks good to me.

shark/iree_utils/compile_utils.py
```diff
@@ -355,11 +350,16 @@ def export_module_to_mlir_file(module, frontend, directory: str):
...
+    function_name,
+    input,
+    config,
+    frontend="torch",
+    send_to_host=True,
 ):
     """Runs a .vmfb file given inputs and config and returns output."""
     device_inputs = [ireert.asdevicearray(config.device, a) for a in input]
-    result = compiled_vm(*device_inputs)
+    result = compiled_vm[function_name](*device_inputs)
```
raikonenfnu, line 362: A stylistic ask, if we know that the functions are limited. i.e `compute_loss`, and `step`, would be nice to have it's own function, since it would look better as `vae.step(arg0)` vs vae("step",arg0)`.

Abhishek-Varma: That's a fair ask and I understand.
Two reasons why, in my opinion, that seems unlikely :-
1. Unlike `forward` (func.func @forward), the accessors generated by IREE is of the following type :-
```
func.func @global$_tensor_constant1$get() -> tensor<f32> {
    %_tensor_constant1 = util.global.load @_tensor_constant1 : tensor<f32>
    return %_tensor_constant1 : tensor<f32>
  }
```
Now invoking the same in Python like, say, `vae.global$_tensor_constant1$()` defeats the "artistic" purpose. :P

2. The number of such functions depends on which model we're dealing with, i.e. how many parameters.

pashu123 (approved):
LGTM!

PR #715, opened 2022-12-23
[SD] Change default VMA large heap block size for windows perf

raikonenfnu:
Windows perform can boost from 2.67s/image to 2.4523s/image. While Linux stays the same.


powderluv (approved):

powderluv:
It could regress 6900xt but let's test

PR #722, opened 2022-12-24
[SD][web] Add device support in the SD web UI

Shukla-Gaurav:
1. Now device selection is available through UI.
2. Models reloading will only happen when there will be a change in the settings(variant + device).

### DIFF
--- a/web/models/stable_diffusion/stable_args.py
+++ b/web/models/stable_diffusion/stable_args.py
@@ -16,7 +16,7 @@ p.add_argument(
 )
 
 p.add_argument(
-    "--negative-prompts",
+    "--negative_prompts",
     nargs="+",
     default=[""],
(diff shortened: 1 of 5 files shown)


web/models/stable_diffusion/stable_args.py
```diff
@@ -32,7 +32,7 @@
 p.add_argument(
     "--seed",
     type=int,
-    default=42,
+    default=-1,
```
raikonenfnu, line 35: Why do we change the seed>

Shukla-Gaurav: actually seed does not matter for web as it is always supplied by UI, but -1 in general means random seed at each generation.

PhaneeshB (approved):
LGTM

PR #723, opened 2022-12-25
[SD] Fix clip inference time

Shukla-Gaurav:
1. Fix clip inference time by adding default warmup_count to 5.
2. It also adds constrained to use tuned models.
3. `model_db.json` is updated with newer custom models.


PhaneeshB (approved):
LGTM

PR #725, opened 2022-12-25
Add support for dreamlike diffusion

PhaneeshB:
- add support for dreamlike diffusion (use `--variant=dreamlike`)
- precision fp32 fp16
- prompt max length - 77


Shukla-Gaurav (approved):
LGTM!

PR #734, opened 2022-12-27
Add a prototype of the model compilation configs for SD

jinchen62:


shark/examples/shark_inference/stable_diffusion/resources/model_db.json
```diff
@@ -64,5 +64,197 @@
...
+            ]
+          }
+        }
+      }
+    },
+    "vae": {
+      "tuned": {
+        "fp16": {
+          "cuda": {
+            "compilation_flags": [
```
qedawkins, line 127: My understanding is that tuned models shouldn't need these flow level transform flags because they're already built into the model. This matches with what was already in opt_params but is a good opportunity to remove these. @yzhang93 can you confirm that we don't need these flags?

qedawkins (changes requested):
Two comments on the json file, otherwise the code changes look good to me.

qedawkins (dismissed):
LGTM! Thanks for helping to clean this up.

qedawkins:
Looks like there's a lint failure. Please run
```
black --line-length=79 .
```

qedawkins (approved):

PR #742, opened 2022-12-30
add generate_sharktank for stable_diffusion model defaults

dan-garvey:
Currently supports 2_1_base models.

### DIFF
--- a/generate_sharktank.py
+++ b/generate_sharktank.py
@@ -18,6 +18,12 @@ import subprocess as sp
 import hashlib
 import numpy as np
 from pathlib import Path
+from shark.examples.shark_inference.stable_diffusion import (
+    model_wrappers as mw,
+)
+from shark.examples.shark_inference.stable_diffusion.stable_args import (
+    args,
+)
 
 
 def create_hash(file_name):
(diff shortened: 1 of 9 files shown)


generate_sharktank.py
```diff
@@ -51,6 +54,59 @@ def save_torch_model(torch_model_list):
...
+                                    precision=precision_value,
+                                    max_len=length,
+                                    width=512,
+                                    height=512,
+                                    use_base_vae=False,
+                                    debug=True,
+                                    sharktank_dir=WORKDIR,
+                                )
+                                args.max_length = length
+                                model_name = f"{args.variant}/{args.version}/{torch_model_name}/{args.precision}/length_{args.max_length}{args.use_tuned}"
```
pashu123, line 104: The model_name should have `width` and `height` params for future uploads.

pashu123: It should also have `batch_size`.

generate_sharktank.py
```diff
@@ -51,6 +54,59 @@ def save_torch_model(torch_model_list):
...
+                from shark.examples.shark_inference.stable_diffusion.stable_args import (
+                    args,
+                )
+
+                args.use_tuned = False
+                args.import_mlir = True
+                args.use_tuned = False
+                args.local_tank_cache = WORKDIR
+                base_sd_versions = ["v1_4", "v2_1"]
+                model_variants = [
```
pashu123, line 67: For clarity: Let's switch to the hf model id's.

shark/examples/shark_inference/stable_diffusion/resources/beta_model_db.json
```diff
@@ -0,0 +1,177 @@
+[
```
pashu123, line 1: Is there any need for de-duplicating this?

pashu123 (changes requested):

generate_sharktank.py
dan-garvey: this is actually dead code, my bad. I'm using the model name from model wrappers, so if we add batch size there it will propagate

shark/examples/shark_inference/stable_diffusion/resources/beta_model_db.json
dan-garvey: My thinking was, when we're ready to replace the old files, it will be easier to simply overwrite the previous file with the beta file, instead of replacing individual entries.

monorimet (approved):
Looks good to me.

PR #761, opened 2023-01-06
SD - Command-line - CKPT update feature

Abhishek-Varma:
This PR contains 7 commits :-

**Commit 1**
-- Adds `use_external_reference` flag to `torch_mlir.compile` to help gather all weight resources.

**Commit 2**
-- Adds checkpoint update feature on Clip, Unet and Vae.

**Commit 3**
-- Adds an assert to check that the current and the new values of weight match in shape.

**Commit 4**
-- Adds two fixes to the CKPT update feature.

**Commit 5**
-- Adds a way to monitor the CKPT update feature which the user can invoke by `show_checkpoint_update` flag.

**Commit 6**
-- Adds README.md entry to use CKPT update feature.

**Commit 7**
-- Fixes couple of lint errors.

This is based on the upstream `torch-mlir` branch which we're using for SHARK -> [cuda_f16](https://github.com/llvm/torch-mlir/tree/cuda_f16) since the patch which would help gather all weight resources outside the `forward` function has been added.

It also needs a temporary fix on IREE :-
1. PR raised on [SHARK-Runtime](https://github.com/nod-ai/SHARK-Runtime/pull/32)
2. Upstream issue being tracked [here](https://github.com/iree-org/iree/issues/11652)

CC: @powderluv @pashu123

### DIFF
--- a/shark/examples/shark_inference/stable_diffusion/README.md
+++ b/shark/examples/shark_inference/stable_diffusion/README.md
@@ -54,3 +54,15 @@ use the flag `--variant=` to specify the model to be used.
 ```shell
 python .\shark\examples\shark_inference\stable_diffusion\main.py --variant=anythingv3 --max_length=77 --prompt="1girl, brown hair, green eyes, colorful, autumn, cumulonimbus clouds, lighting, blue sky, falling leaves, garden"
 ```
+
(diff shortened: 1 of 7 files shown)


powderluv:
deprecated.

PR #769, opened 2023-01-09
Re-enable TF GPU device

mariecwhite:


mariecwhite:
This change causes TF CPU to stall. Will look into it.

mariecwhite:
This is now working. PTAL

mariecwhite:
Thanks for the explanation.

For 1, we would prefer to have accuracy verified for each benchmark run so decoupling is not ideal.
For 2, did you look into having the process return results? Something like this: https://stackoverflow.com/questions/10415028/how-to-get-the-return-value-of-a-function-passed-to-multiprocessing-process

PR #790, opened 2023-01-10
add json data for each image

PhaneeshB:
- adds json info for each generated image with the same name as the image file
- changes images name to take partial prompt, seed and date time.
- gives option to save images in a specified dir using `--output_dir` flag

### DIFF
--- a/shark/examples/shark_inference/stable_diffusion/stable_args.py
+++ b/shark/examples/shark_inference/stable_diffusion/stable_args.py
@@ -123,6 +123,12 @@ p.add_argument(
     help="other supported schedulers are [PNDM, DDIM, LMSDiscrete, EulerDiscrete, DPMSolverMultistep]",
 )
 
+p.add_argument(
+    "--output_dir",
+    type=str,
+    default=None,
+    help="Directory path to save the output images and json",
+)
 ##############################################################################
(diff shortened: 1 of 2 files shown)


shark/examples/shark_inference/stable_diffusion/stable_args.py
```diff
@@ -123,6 +123,12 @@ def path_expand(s):
     help="other supported schedulers are [PNDM, DDIM, LMSDiscrete, EulerDiscrete, DPMSolverMultistep]",
 )
 
+p.add_argument(
+    "--output_dir",
+    type=str,
+    default=None,
```
powderluv, line 129: should this be `pwd`

PhaneeshB: yes, it is pwd or the path specified by the user.
taken into consideration in main.py
also it will create a path specified by user if not already created

powderluv:
Please test this will the exe. We may have to add the new imports to the spec file.

PhaneeshB:
> Please test this will the exe. We may have to add the new imports to the spec file.

This should be good to go as these changes don't affect web just cli
@Shukla-Gaurav

Shukla-Gaurav (approved):
LGTM!

PR #801, opened 2023-01-12
Increase JPEG output quality & disable subsampling

xzuyn:
Increased to JPEG95 from the default JPEG75 which is way too compressed. Output image size is now ~100-200kb. Previously was ~20kb.

### DIFF
--- a/shark/examples/shark_inference/stable_diffusion/main.py
+++ b/shark/examples/shark_inference/stable_diffusion/main.py
@@ -273,6 +273,8 @@ if __name__ == "__main__":
         }
         prompt_slice = re.sub("[^a-zA-Z0-9]", "_", args.prompts[i][:15])
         img_name = f"{prompt_slice}_{args.seed}_{i}_{dt.now().strftime('%y%m%d_%H%M%S')}"
-        pil_images[i].save(output_path / f"{img_name}.jpg", quality=95, subsampling=0)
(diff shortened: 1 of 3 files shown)


powderluv (commented):
Awesome. Do we need it for the cli too ?

xzuyn:
I've only really used the GUI so I'm not sure. I think that should add the change for cli.

powderluv (approved):

powderluv:
Please run black --line-length 79 .

You can install with pip install black.

xzuyn:
what for?

powderluv:
Oh the CI has a bunch of python code formatting checks. See the build failure in the CI.

xzuyn:
fixed the line length formatting.

powderluv (approved):
We can merge this -- but don't we just want 100%

xzuyn:
Documentation says going above 95 isn't recommended.

> The image quality, on a scale from 0 (worst) to 95 (best), or the string keep. The default is 75. Values above 95 should be avoided; 100 disables portions of the JPEG compression algorithm, and results in large files with hardly any gain in image quality. The value keep is only valid for JPEG files and will retain the original image quality level, subsampling, and qtables.

Maybe a toggle between PNG and JPEG95 at some point?

powderluv:
yup reasonable to add png

PR #805, opened 2023-01-12
small fixes

PhaneeshB:
- change mlir module name in torch utils
- add extra_args as a keyword arg in vulkan utils


pashu123 (approved):
Thanks!

PR #806, opened 2023-01-12
[SD] Reorganize the stable diffusion model.

Shukla-Gaurav:
The stable diffusion codebase has been reorganized to make it more modular so that the same script can be used for web as well as cli, instead of duplicating the whole codebase.


Shukla-Gaurav:
This draft PR is an effort to re-organize the stable diffusion codebase. This should make the code more modular and the process of adding new features like img2img, upscalar less redundant. 
I would love to get feedbacks/suggestions on this restructuring. @powderluv @qedawkins @raikonenfnu  @pashu123 @PhaneeshB

raikonenfnu:
Some ideas/notes:

-Perhaps we should add a new directory add the root called "apps" and this directory can go in this a.k.a apps/stable_diffusion. 
-Is upscaler not in this patch?

qedawkins:
+1 on the apps idea. Or rather, it would be good to understand what the intended import graph looks like for stable diffusion. Likely needs to be coordinated with #742 as well, as the model wrappers are moved with this change to a location that can't easily be imported by another module (I think the solution is probably just to put model_wrappers.py in the tank then and import from there?)

stablediffusion/utils/stable_diffusion.py
```diff
@@ -0,0 +1,235 @@
+import torch
```
pashu123, line 1: Can create a different folder namely - model_pipeline and add `shark_stable_diffusion_pipeline.py`.

Shukla-Gaurav:
> Some ideas/notes:
> 
> -Perhaps we should add a new directory add the root called "apps" and this directory can go in this a.k.a apps/stable_diffusion. -Is upscaler not in this patch?

Adding the features incrementally. After img2img, I will add upscaler in the next patch.

powderluv:
Please readd model.json 

my testing it seems good.

powderluv (approved):

powderluv (approved):

PR #812, opened 2023-01-13
add single execution multi-run SD

PhaneeshB:
- we can use flag `--runs` to specify the number of images to be generated in one execution.
- this will use all the same prompt and generate `runs` number of images by consecutively running SD with a random seed value for each run.
- if `--seed` is provided in cli it will only be used for the first image.
- all images will be saved in current working directory or path specified by `--output_dir`


PhaneeshB:
This is a temporary addition to support multiple runs and will be modified with the refactored SD.
@Shukla-Gaurav

powderluv (approved):

powderluv (approved):

PR #819, opened 2023-01-16
[SD-CLI] Add CKPT support to update models irrespective of `import_mir` flag

Abhishek-Varma:
-- This commit adds CKPT support to update models irrespective of `import_mlir` flag.

### DIFF
--- a/shark/examples/shark_inference/stable_diffusion/README.md
+++ b/shark/examples/shark_inference/stable_diffusion/README.md
@@ -54,3 +54,16 @@ use the flag `--variant=` to specify the model to be used.
 ```shell
 python .\shark\examples\shark_inference\stable_diffusion\main.py --variant=anythingv3 --max_length=77 --prompt="1girl, brown hair, green eyes, colorful, autumn, cumulonimbus clouds, lighting, blue sky, falling leaves, garden"
 ```
+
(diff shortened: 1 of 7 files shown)


pashu123:
@Abhishek-Varma 1. We will have to version on which this .ckpt is based on i.e. v1.4 or v2.1 along with .ckpt. 
2. Also, we don't need separate .ckpt for unet, vae and clip rather we should just provide one ckpt we will enumerate and update all the models. https://huggingface.co/stabilityai/stable-diffusion-2/tree/main

pashu123 (approved):

Abhishek-Varma:
I've tested this on the following variants' CKPT :-
1. CompVis/stable-diffusion-v1-4
2. Linaqruf/anything-v3.0
3. wavymulder/Analog-Diffusion
4. prompthero/openjourney
5. andite/anything-v4.0

And I confirm it works.

I used the `.ckpt` files to generate the structure for `.bin` files just what we for diffusers using [convert_original_stable_diffusion_to_diffusers.py](https://github.com/huggingface/diffusers/blob/main/scripts/convert_original_stable_diffusion_to_diffusers.py).

Once done, I was able to use that to generate/compile custom models using `custom_model` flag.

NOTE: User can also directly provide the huggingface repo-id using `custom_model` flag now.

PR #832, opened 2023-01-18
Refactor shark_tank models and custom models.

pashu123:
The custom models shouldn't depend on shark_tank in anyway.


Abhishek-Varma (approved):
LGTM!

PR #849, opened 2023-01-20
Modify the default for --hf_model_id flag

yzhang93:


pashu123 (approved):

Abhishek-Varma:
Hi @yzhang93 @pashu123 

This is for my understanding.
Why was this change required explicitly instead of directly updating `hf_model_id` during the run itself?

pashu123:
> Hi @yzhang93 @pashu123
> 
> This is for my understanding. Why was this change required explicitly instead of directly updating `hf_model_id` during the run itself?

Just making it as the default model. If someone doesn't pass hf_model_id flag.

Abhishek-Varma:
This should've been complemented with an update to README.md as well since it refers to [CompVis/stable-diffusion-v1-4](https://huggingface.co/CompVis/stable-diffusion-v1-4) as the default value of `hf_model_id`.

And it seems like majority of the models in `variants.json` are based on [CompVis/stable-diffusion-v1-4](https://huggingface.co/CompVis/stable-diffusion-v1-4).

Any specific reason to make `stabilityai/stable-diffusion-2-1-base` as the default one? @pashu123 

Since majority of the models depend on the other one, it only makes sense for us to revert back to the `v1_4`.

yzhang93:
> This should've been complemented with an update to README.md as well since it refers to [CompVis/stable-diffusion-v1-4](https://huggingface.co/CompVis/stable-diffusion-v1-4) as the default value of `hf_model_id`.
> 
> And it seems like majority of the models in `variants.json` are based on [CompVis/stable-diffusion-v1-4](https://huggingface.co/CompVis/stable-diffusion-v1-4).
> 
> Any specific reason to make `stabilityai/stable-diffusion-2-1-base` as the default one? @pashu123
> 
> Since majority of the models depend on the other one, it only makes sense for us to revert back to the `v1_4`.

The only reason is v2-1base has the highest performance, and it's the model we used for our demos.

PR #851, opened 2023-01-21
Collapsibles for Win10 and Linux users

erkinalp:


powderluv:
thanks for the clean up

powderluv (approved):

PR #863, opened 2023-01-24
[SD-CLI] Make using `ckpt_loc` and `hf_model_id` easier

Abhishek-Varma:
-- Currently we require users to specify the base model on which the custom
   model (.ckpt) is tuned on. Even for running a HuggingFace repo-id, we
   require the users to go a tedious way of adding things to variants.json.

-- This commit aims to address the above issues and will be treated as a
   starting point for a series of design changes which makes using SHARK's SD
   easier.


pashu123 (dismissed):
LGTM!

pashu123 (approved):

PR #869, opened 2023-01-25
Add SD model annotation on fly

yzhang93:
Annotation on fly function only works when this PR merged upstream https://github.com/llvm/torch-mlir/pull/1797.
Before this patch merged, we'll need to make sure `--no-use_tuned` is passed when `--import_mlir` is used. The default of `--import_mlir` is set as `False` for now.

### DIFF
--- a/shark/examples/shark_inference/stable_diffusion/model_wrappers.py
+++ b/shark/examples/shark_inference/stable_diffusion/model_wrappers.py
@@ -3,7 +3,6 @@ from transformers import CLIPTextModel
 from utils import compile_through_fx, get_opt_flags
 from resources import base_models
 from collections import defaultdict
-from sd_annotation import tuned_compile_through_fx
 import torch
 import sys
 
@@ -131,22 +130,14 @@ class SharkifyStableDiffusionModel:
         inputs = tuple(self.inputs["vae"])
         is_f16 = True if self.precision == "fp16" else False
         vae_name = "base_vae" if self.base_vae else "vae"
-        if self.use_tuned:
-            shark_vae = tuned_compile_through_fx(
-                vae,
-                inputs,
-                is_f16=is_f16,
-                model_name=vae_name + self.model_name + "_tuned",
-                extra_args=get_opt_flags("vae", precision=self.precision),
-            )
-        else:
-            shark_vae = compile_through_fx(
-                vae,
-                inputs,
-                is_f16=is_f16,
-                model_name=vae_name + self.model_name,
-                extra_args=get_opt_flags("vae", precision=self.precision),
-            )
+        shark_vae = compile_through_fx(
+            vae,
+            inputs,
+            is_f16=is_f16,
+            model_name=vae_name + self.model_name,
+            use_tuned=self.use_tuned,
+            extra_args=get_opt_flags("vae", precision=self.precision),
+        )
         return shark_vae
 
     def get_unet(self):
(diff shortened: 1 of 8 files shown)


shark/torch_mlir_utils.py
```diff
@@ -56,6 +56,7 @@ def get_torch_mlir_module(
     input: tuple,
     dynamic: bool,
     jit_trace: bool,
+    return_str: bool,
```
pashu123, line 59: Let's set this parameter default to `False`. Otherwise, users have to specify it everywhere.

shark/examples/shark_inference/stable_diffusion/sd_annotation.py
```diff
@@ -116,7 +139,85 @@
...
+        use_winograd = False
+        lowering_config_dir = load_lower_configs()
+        tuned_model, output_path = annotate_with_lower_configs(
+            mlir_model, lowering_config_dir, model_name, use_winograd
+        )
     print(f"Saved the annotated mlir in {output_path}.")
+    return tuned_model, output_path
+
+
+def tuned_compile_through_fx(
```
pashu123, line 180: Can we merge both the function, i.e. `compile_through_fx` and `tuned_compile_through_fx` with an extra parameter `tuning` set defaulted to False?

shark/examples/shark_inference/stable_diffusion/model_wrappers.py
```diff
@@ -166,14 +178,24 @@ def forward(
...
-        shark_unet = compile_through_fx(
-            unet,
-            inputs,
-            model_name="unet" + self.model_name,
-            is_f16=is_f16,
-            f16_input_mask=input_mask,
-            extra_args=get_opt_flags("unet", precision=self.precision),
-        )
+        if self.use_tuned:
+            shark_unet = tuned_compile_through_fx(
```
pashu123, line 182: Please merge both the function—Handling control flow can be done in `utils.py`.

pashu123 (changes requested):

shark/torch_mlir_utils.py
yzhang93: Done.

shark/examples/shark_inference/stable_diffusion/sd_annotation.py
yzhang93: I've merged the two functions in utils.py.

shark/examples/shark_inference/stable_diffusion/utils.py
```diff
@@ -70,12 +72,40 @@ def compile_through_fx(
...
+            if "cuda" in args.device:
+                output_path = (
+                    f"{args.annotation_output}/{model_name}_orig.mlir"
+                )
+                with open(output_path, "w") as f:
+                    f.write(mlir_module)
+                    f.close()
+                mlir_module = output_path
+
+            tuned_model, tuned_model_path = sd_model_annotation(
```
pashu123, line 99: Don't you think we will have to check for `batch_size` `height` and `width` and `max_len` too when applying annotation?

yzhang93: Good point! I only tested the codes with the default settings on vulkan. And the tuned configs only apply to batch_size 1, height/width 512. The max length could be 77 or 64 though. The goal of this PR is making the basic functionality working for the default settings. I will test and add more settings in the following patches.

pashu123: Cool! We can do `sys.exit("Tuned models are currently supported for ...")` if the above criteria aren't met.

yzhang93: I think it should fall into untuned models if the tuned models are not available, similar to we used to check for device.

pashu123 (approved):
Thanks! LGTM.

PR #873, opened 2023-01-25
[SD-CLI] Add `batch_size` command-line arg + prompt processing

Abhishek-Varma:
-- This commit adds `batch_size` command-line arg.
-- It also involves replicating the prompt `batch_size` no. of times.

NOTE:

1. If user specifies just `1` prompt along with `batch_size`, then we'll be using the same prompt for inferring.
2. Else if user provides `>1` prompt then we'd need the number of prompts to be `=batch_size`. This gives the user a space to introduce different prompts and see them all getting inferred.
3. `batch_size` has been restricted to stay between `[1, 3]` because beyond that it is currently generating garbage output.
4. So, we'd require `runs` flag in case user wants to generate some arbitrary number of images (`>3`).


pashu123 (approved):
LGTM!

PR #876, opened 2023-01-26
Adds pytest-forked dependency to fix pytest memory accumulation issues.

monorimet:
This adds pytest-forked to requirements.txt -- to use this plugin, simply add `--forked` to your pytest command.
This way, each individual test case is run in its own subprocess, and error handling is seamlessly piped from the subprocess to pytest's reports.

@mariecwhite this should unblock us on the TF GPU enablement.

### DIFF
--- a/.github/workflows/test-models.yml
+++ b/.github/workflows/test-models.yml
@@ -100,9 +100,9 @@ jobs:
       if: matrix.suite == 'cpu'
       run: |
         cd $GITHUB_WORKSPACE
-        PYTHON=python${{ matrix.python-version }} BENCHMARK=1 IMPORTER=1 ./setup_venv.sh
+        PYTHON=python${{ matrix.python-version }} IMPORTER=1 ./setup_venv.sh
         source shark.venv/bin/activate
-        pytest --benchmark --ci --ci_sha=${SHORT_SHA} -s --local_tank_cache="/data/anush/shark_cache" tank/test_models.py -k cpu --update_tank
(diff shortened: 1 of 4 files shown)


.github/workflows/test-models.yml
```diff
@@ -100,9 +100,9 @@ jobs:
       if: matrix.suite == 'cpu'
       run: |
         cd $GITHUB_WORKSPACE
-        PYTHON=python${{ matrix.python-version }} BENCHMARK=1 IMPORTER=1 ./setup_venv.sh
```
dan-garvey, line 103: is this no longer required

.github/workflows/test-models.yml
```diff
@@ -126,12 +126,12 @@ jobs:
         export DYLD_LIBRARY_PATH=/usr/local/lib/
         echo $PATH
         pip list | grep -E "torch|iree"
-        pytest -s --ci --ci_sha=${SHORT_SHA} --local_tank_cache="/Volumes/builder/anush/shark_cache" tank/test_models.py -k vulkan --update_tank
+        pytest --ci --ci_sha=${SHORT_SHA} --local_tank_cache="/Volumes/builder/anush/shark_cache" -k vulkan --update_tank
```
dan-garvey, line 129: why does this work without a target for pytest?

dan-garvey (approved):
check comments are ok then merge

.github/workflows/test-models.yml
monorimet: tank/test_models.py is the only valid target from the base SHARK directory -- the other ones are ignored via SHARK/pytest.ini

PR #889, opened 2023-01-30
adding more robust main.py testing

dan-garvey:
Currently disabled for CUDA because it only produces black images on my vm and on builder. 

Adds support for any model listed in model_config.json. However, any additions to model_config.json will need to be added to the golden images in order to work correctly. Alternatively I can change this to fail the builder if a new golden value isn't added.

Stable diff v1 also appears to be broken on ToM

Any failures in image comparison will be automatically uploaded to gs://shark_tank/testdata/builder
Note that any file 842b is a black image.  (failure)


powderluv:
Let's fix v1 and the CUDA builds and expand the testing. Ok to fail build if golden image is missing. Just suggest the remedy

PR #894, opened 2023-01-31
Added flags for metadata information.

EvanGuanSF:
Two new flags for the stable diffusion program:
-- Added `--save_metadata_to_json` flag (default `True`): whether or not to write a generation data .json file along with every generated image.
-- Added `--write_metadata_to_png` flag (default `False`): whether or not to write generation data into PNG text chunk area for every generated image png.


powderluv (changes requested):
You also have to add the flag in the webui in the [web/ directory](https://github.com/nod-ai/SHARK/blob/8111f8bf35afb9f0ca54a36b6bd64b866952d8b0/web/models/stable_diffusion/main.py#L94). (That should merge with main.py code soon)

EvanGuanSF:
Forgot to remove some old code, will push clean version momentarily.

powderluv:
please run the black linter (`pip install black`)  and then `black --line-length 79 .`

web/models/stable_diffusion/main.py
```diff
@@ -92,11 +92,28 @@ def save_output_img(output_img):
         )
     else:
         out_img_path = Path(generated_imgs_path, f"{out_img_name}.png")
-        output_img.save(out_img_path, "PNG")
+        pngInfo = PngImagePlugin.PngInfo()
+
+        if args.write_metadata_to_png:
+            # set height and width.
+            height = 512  # default height of Stable Diffusion
+            width = 512  # default width of Stable Diffusion
```
powderluv, line 100: I guess eventually we will need to set these based on the chosen width/height.

powderluv (dismissed):

web/models/stable_diffusion/main.py
EvanGuanSF: Indeed, it felt a little awkward copy pasting hard coded values. 😅

powderluv:
Height / Width support and a major refactor just landed with https://github.com/nod-ai/SHARK/commit/c124b76328e5470f1601093e90258d1887c7fdf8 .  We will need to rebase onto that patch in the new location since these two files web/index.py and main.py are now deprecated. Sorry about the timing with your PR, please let me know if you can rebase or we can help with it

EvanGuanSF:
> Height / Width support and a major refactor just landed with [c124b76](https://github.com/nod-ai/SHARK/commit/c124b76328e5470f1601093e90258d1887c7fdf8) . We will need to rebase onto that patch in the new location since these two files web/index.py and main.py are now deprecated. Sorry about the timing with your PR, please let me know if you can rebase or we can help with it

This is my first time doing anything like this but I'll give it a shot. I'll let you know if I need someone else to take it over.

EvanGuanSF:
Fixed the old deprecated stuff (I think), looking into /apps now.

powderluv (approved):

apps/stable_diffusion/src/utils/stable_args.py
```diff
@@ -270,6 +270,20 @@ def path_expand(s):
     help="flag to clear all mlir and vmfb from common locations. Recompiling will take several minutes",
 )
 
+p.add_argument(
+    "--save_metadata_to_json",
+    default=False,
```
powderluv, line 275: Don't we want this to be true ? Or if this is false we should toggle the png metadata to true.

EvanGuanSF: I am honestly not sure which is preferred here. As far as I can tell the `app` and `web` versions are already writing metadata to a csv? I have been having trouble getting both GUI versions to run consistently on my machine (somehow 32GB does not seem like enough ram and I keep getting memory errors).

EvanGuanSF: For clarification, my initial intention and use case for these flags was for the CLI version of the program in `shark/examples/shark_inference/stable_diffusion`. That version does not write to .csv and defaults to a json file output.

Personally I like the idea of writing metadata to the png itself by default. It keeps the data with the image and the extra space cost is maybe only 1-2KB and is far cleaner in file browsers.

powderluv: Yeah I like it to be in the meta data too. I'm testing the new flow and the imports. Once that is good we can land your change and iterate on the right defaults.

EvanGuanSF:
Sorry for the constant small commits, I've finally got the apps gui working reliably (needed to drop to 384x384) so I could test everything and it all seems to be good to go now. The gui produces proper outputs for the flags. There are other problems to deal with but I think they are outside the scope of what I want to do for now. 

TODO:
-- output directory in the apps gui is being ignored for the images and json files. The gui text value `output_dir` is currently not sent as an arg for the backend txt2img.py.
-- Image seed number being stored in metadata is incorrect in the case that the seed being given by the gui is outside the range of a uint32. The program randomly generates a seed in those cases but that seed is not relayed to the `args.seed` variable which is used to populate the metadata.

powderluv (approved):

PR #899, opened 2023-01-31
fixed a bug that would sometimes cause intel-gpu to appear unsupported

Eliasj42:
Fixed a bug that would cause shark to say level_zero is unsupported when trying to specify which gpu to use

Also removed code that had been commented out


monorimet (approved):
LGTM! good catch

PR #907, opened 2023-02-01
Add PyInstaller for apps/

powderluv:
Build with pyinstaller.exe .\apps\stable_diffusion\web\shark_sd.spec

normal flow works. exe is missing a few json files


Shukla-Gaurav:
Looks good to me. 
There are 2 more files with relative imports:
`scripts/__init__.py and schedulers/sd_schedulers.py`.

powderluv:
Verified both web and cli. Landing.

PR #908, opened 2023-02-01
[APPS-SD] Fix a few bugs and bring it up to speed with SD CLI

Abhishek-Varma:
-- This commit fixes a few bugs in `apps`'s SD CLI and brings it
   up to speed with SD CLI.

### DIFF
--- a/apps/stable_diffusion/src/models/__init__.py
+++ b/apps/stable_diffusion/src/models/__init__.py
@@ -6,4 +6,6 @@ from apps.stable_diffusion.src.models.opt_params import (
     get_unet,
     get_clip,
     get_tokenizer,
+    get_params,
+    get_variant_version,
 )
--- a/apps/stable_diffusion/src/models/opt_params.py
+++ b/apps/stable_diffusion/src/models/opt_params.py
@@ -14,6 +14,10 @@ hf_model_variant_map = {
 }
 
 
+def get_variant_version(hf_model_id):
+    return hf_model_variant_map[hf_model_id]
+
+
 def get_params(bucket_key, model_key, model, is_tuned, precision):
     iree_flags = []
     if len(args.iree_vulkan_target_triple) > 0:
@@ -60,7 +64,7 @@ def get_params(bucket_key, model_key, model, is_tuned, precision):
 
 
 def get_unet():
-    variant, version = hf_model_variant_map[args.hf_model_id]
+    variant, version = get_variant_version(args.hf_model_id)
     # Tuned model is present only for `fp16` precision.
     is_tuned = "tuned" if args.use_tuned else "untuned"
     if "vulkan" not in args.device and args.use_tuned:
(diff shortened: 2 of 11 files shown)


powderluv (approved):
Please check the webui works too. @Shukla-Gaurav  please take a look. I'll rebase my patch on this

apps/stable_diffusion/src/models/opt_params.py
```diff
@@ -13,6 +16,14 @@
...
 }
 
+# TODO: Implement a better way to handle the KeyNotFound error by delaying the invocation.
+variant = None
+version = None
+try:
+    variant, version = hf_model_variant_map[args.hf_model_id]
+except:
+    pass
+
```
Shukla-Gaurav, line 26: This has to be local to each function, i.e., means inside get_unet(), get_vae() and get_clip(), as these functions can be called multiple times with different model ids.

apps/stable_diffusion/src/models/opt_params.py
```diff
@@ -60,7 +71,6 @@ def get_params(bucket_key, model_key, model, is_tuned, precision):
 
 
 def get_unet():
-    variant, version = hf_model_variant_map[args.hf_model_id]
```
Shukla-Gaurav, line 63: This is necessary to not fix the version and variant when executing it multiple times.

apps/stable_diffusion/src/models/__init__.py
```diff
@@ -1,2 +1,10 @@
 from .model_wrappers import SharkifyStableDiffusionModel
-from .opt_params import get_vae, get_unet, get_clip, get_tokenizer
+from .opt_params import (
+    get_vae,
+    get_unet,
+    get_clip,
+    get_tokenizer,
+    get_params,
+    version,
+    variant,
```
Shukla-Gaurav, line 9: We should avoid such global variables(as they change with multiple calls) and importing them. As with web, there will be multiple calls with changes to model id's.

apps/stable_diffusion/src/utils/sd_annotation.py
```diff
@@ -0,0 +1,191 @@
...
+
+
+device = (
+    args.device if "://" not in args.device else args.device.split("://")[0]
+)
+
+
+# Download the model (Unet or VAE fp16) from shark_tank
+def load_model_from_tank():
+    from ..models import get_params, version, variant
```
Shukla-Gaurav, line 20: Please fix relative imports. They have problem with pyinstaller.

apps/stable_diffusion/src/utils/sd_annotation.py
```diff
@@ -0,0 +1,191 @@
...
+    download_public_file,
+    WORKDIR,
+)
+from shark.parser import shark_args
+from .stable_args import args
+
+
+device = (
+    args.device if "://" not in args.device else args.device.split("://")[0]
+)
```
Shukla-Gaurav, line 15: avoid such global variables.

Shukla-Gaurav (changes requested):
Thanks, needs following changes.

apps/stable_diffusion/src/utils/sd_annotation.py
dan-garvey: agree here and elsewhere with what gaurav said. We want to move away from global variables and dependence on args outside of the script that is called from the command line

Shukla-Gaurav (approved):
Thanks, LGTM!

PR #911, opened 2023-02-01
Drop old cli and webui

powderluv:


dan-garvey:
paths in generate_sharktank imports and build_tools/stable_diffusion_testing need to be edited

dan-garvey (approved):

PR #919, opened 2023-02-02
disable Torch Inductor autotuner in benchmarks

monorimet:
Torch Inductor currently fails with a CUDA error all torchvision models and some bert models.


powderluv (approved):

powderluv:
I think this crashed one of the a100 machines. @mariecwhite fyi

Issue #582, opened 2022-12-07
[WEB]Make the dark theme default while launching web server


Shukla-Gaurav:

powderluv:
done

PR #954, opened 2023-02-05
Make the dark theme default while launching web server

drumicube:
This PR address issue https://github.com/nod-ai/SHARK/issues/582

Sadly, I haven't found a simple way to pass a theme as argument during Gradio init, to my knowledge this isn't supported.
Digging into their theming  system, when using http://localhost:8080/?__theme=dark , I realised they simply surcharges :root with a .dark class to apply the selected theme.

Using our custom sd_dark_theme.css, I copied all their .dark content to :root, as a result this makes the default Gradio white theme behave exactly the same way as the dark one.

I have also sanitized sd_dark_theme.css to replace all hardcoded colors by appropriate color vars:
Theme customisations should now only be needed by modifying vars declared in :root
I have tested this by commenting all :root rules, and indeed, minus the white hardcoded text logo, the white theme was displayed correctly.

### DIFF
--- a/apps/stable_diffusion/web/css/sd_dark_theme.css
+++ b/apps/stable_diffusion/web/css/sd_dark_theme.css
@@ -1,5 +1,153 @@
+
+/* Overwrite the Gradio default theme with their .dark theme declarations */
+
+:root {
+    --color-focus-primary: var(--color-grey-700);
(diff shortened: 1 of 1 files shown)


powderluv:
Thank you this is awesome!!!

powderluv:
tested and looks great. Thank you

PR #958, opened 2023-02-06
rand seed generation and check

PhaneeshB:
This is a temporary fix to get the correct seed info while saving the images.
Ideally this should be part of the pipeline and will be integrated in upcoming PRs.


powderluv:
let try to fix it the right way. no need to rush it.

Shukla-Gaurav:
This has been fixed, closing it.

PR #966, opened 2023-02-07
generate sharktank for apps dir

dan-garvey:
also adds support for the sub-models


apps/stable_diffusion/src/utils/stable_args.py
```diff
@@ -341,5 +349,37 @@ def path_expand(s):
     action=argparse.BooleanOptionalAction,
     help="Apply Winograd on selected conv ops.",
 )
+##############################################################################
```
monorimet, line 352: These are still used for SD? Can we remove them? They aren't used in generate_sharktank.py anymore

monorimet (approved):
Everything looks good to me, should we test nightly workflow with this patch before merging?

powderluv:
Can we please land this ? And other in-flight patches

yzhang93 (approved):

PR #967, opened 2023-02-07
Update .gitignore to exclude models

cstueckrath:
the models folder will be stashed along with other changes and most likely kill git doing so.


powderluv (approved):

powderluv:
thank you

PR #968, opened 2023-02-07
Fix restart SD session error + override args.use_tuned temporarily

Abhishek-Varma:
-- This commit fixes the session restart error for SD.
-- It also overrides `args.use_tuned` for `import_mlir`, and sets `use_tuned` as `False`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -176,5 +176,9 @@ onnx_models/
 # Generated images
 generated_imgs/
 
+# Custom model related artefacts
+apps/stable_diffusion/src/utils/resources/variants.json
+models/
+
 # models folder
 apps/stable_diffusion/web/models/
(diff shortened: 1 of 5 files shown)


.gitignore
```diff
@@ -176,5 +176,9 @@ onnx_models/
 # Generated images
 generated_imgs/
 
+# Custom model related artefacts
+apps/stable_diffusion/src/utils/resources/variants.json
```
powderluv, line 180: Do we generate this file ?

powderluv (approved):
This is ok for now but we will need to apple tuning on the fly soon. @yzhang93

.gitignore
Abhishek-Varma: Yes, Anush.
This is used to serve us two purpose :-
1. Fetch hf_model_id once the session starts - just like we fetch already compiled vmfbs.
2. Fetch hf_model_id to bypass the retry mechanism in case the vmfbs are deleted - saves the user the turnaround time involved in retry method. Therefore even if we use `--clear_all`, we can directly use the knowledge stored in this file and compile the sub-modules directly without any hassle.

PR #971, opened 2023-02-07
Add attention slicing to stable diffusion txt2img

gpetters94:
It produces good images, not clear if it's improving performance.

### DIFF
--- a/apps/stable_diffusion/src/models/model_wrappers.py
+++ b/apps/stable_diffusion/src/models/model_wrappers.py
@@ -226,6 +226,11 @@ class SharkifyStableDiffusionModel:
                 )
                 self.in_channels = self.unet.in_channels
                 self.train(False)
+                if(args.attention_slicing is not None and args.attention_slicing != "none"):
+                    if(args.attention_slicing.isdigit()):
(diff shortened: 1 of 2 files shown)


apps/stable_diffusion/scripts/txt2img.py
```diff
@@ -224,6 +224,7 @@ def txt2img_inf(
             args.use_base_vae,
             args.use_tuned,
         )
+        txt2img_obj.enable_attention_slicing("auto")
```
powderluv, line 227: can we please add it as an arg.attention_slicing and set it to auto / max ? You will also need to add it to the web/index.py.

pashu123:
@gpetters94 The models are not present in the pipeline, they are created in another file https://github.com/nod-ai/SHARK/blob/main/apps/stable_diffusion/src/models/model_wrappers.py and used from there. Did you see any performance gaps or new ops?

pashu123:
> @gpetters94 The models are not present in the pipeline, they are created in another file https://github.com/nod-ai/SHARK/blob/main/apps/stable_diffusion/src/models/model_wrappers.py and used from there so the attention slicing won't work. Did you see any performance gaps or new ops?

PR #975, opened 2023-02-08
Fix SD restart error in exe file

Abhishek-Varma:
-- This commit fixes SD restart error in exe file by creating variants.json in CWD instead of a relative path.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -177,7 +177,7 @@ onnx_models/
 generated_imgs/
 
 # Custom model related artefacts
-apps/stable_diffusion/src/utils/resources/variants.json
+variants.json
 models/
 
 # models folder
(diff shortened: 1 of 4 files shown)


powderluv:
Hold on but no one has reported an issue

powderluv (approved):
Can you please test the exe to see if you notice it. We can wait for one use to report it before landing this

Abhishek-Varma:
> Can you please test the exe to see if you notice it. We can wait for one use to report it before landing this

I performed the following test on the 506 exe of WebUI and was able to repro the issue :-
1. Run a custom checkpoint model.
2. Close the WebUI.
3. Relaunch it and use the same checkpoint model - leads to the base model config missing error.

Then worked on my patch followed by again testing it on a combination of ways on both web/cli normal as well as exe execution. I confirm that with this patch the issue gets fixed.

Let's wait for one user to report the issue and we can land this thereafter.

powderluv:
Cool let's land it.

PR #978, opened 2023-02-08
Delete unnecessary arg setting

yzhang93:


powderluv (approved):

powderluv:
how does the webui handle this ? Would it always assume it can tune ? What if there tuning is not available for a model ?

PR #980, opened 2023-02-08
Modify the annotation OTF to return bytecode module

yzhang93:


apps/stable_diffusion/src/utils/utils.py
```diff
@@ -253,11 +246,7 @@ def set_init_device_flags():
     ):
         args.use_tuned = False
 
-    elif "cuda" in args.device and get_cuda_sm_cc() not in [
-        "sm_80",
-        "sm_84",
-        "sm_86",
-    ]:
+    elif "cuda" in args.device and get_cuda_sm_cc() not in ["sm_80"]:
```
dan-garvey, line 249: what is this change for?

dan-garvey (approved):
Looks great, I should be able to generate with this!

apps/stable_diffusion/src/utils/utils.py
yzhang93: We only want the cuda tuned configs to apply on A100, as we don't have tuning configs tested on other devices. We had seen problems applying the same config on sm_89. sm_89 has already been removed from the list.

PR #981, opened 2023-02-09
Fix local_tank_cache handling in shark_downloader.

monorimet:
On import of the shark downloader, the shark_tank local cache (`WORKDIR` in shark_downloader.py) should be determined by the following method.

1. If shark_args.local_tank_cache is set, `WORKDIR` is set to that path.
2. If shark_args.local_tank_cache is not specified and there exists a locally generated shark_tank (at SHARK/gen_shark_tank), use that directory as `WORKDIR`. This is mostly for the nightly CI.
3. If conditions for 1 and 2 are not met, use `<HOME>/.local/shark_tank/` as `WORKDIR`.

Option #2's condition was written as an `if` statement instead of `elif` which caused option #3 to overwrite WORKDIR.

Ideally we don't use shark_args at all, but that is a larger undertaking.

### DIFF
--- a/.github/workflows/test-models.yml
+++ b/.github/workflows/test-models.yml
@@ -136,7 +136,7 @@ jobs:
         export DYLD_LIBRARY_PATH=/usr/local/lib/
         echo $PATH
         pip list | grep -E "torch|iree"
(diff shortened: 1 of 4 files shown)


dan-garvey (approved):

dan-garvey:
> Ideally we don't use shark_args at all, but that is a larger undertaking.

👏

PR #982, opened 2023-02-09
WebUi: Add support for more command arguments

drumicube:
Many command arguments are not supported in the webUi for now (width, height, prompt, max_length, etc.), they are forced to a static value declared in apps/stable_diffusion/web/index.py

Each time I launch the WebUi, I'm loosing time bringing back my favorite settings.
This PR try to address this issue:

All WebUi fields expect the **Device** can now be initialized at launch using the proper command argument.
Support was added in the WebUi for those arguments:
```
--hf_model_id, --prompts, --negative_prompts, --steps, --seed, --batch_size, --height, --width, --guidance_scale, --max_length, --precision, --scheduler, --batch_count, --batch_size, --ckpt_loc, --hf_model_id, --save_metadata_to_json, --write_metadata_to_png
```

* If a ckpt_loc is provided and available in the custom model dropdown , it should also be selected by default.
* I took the liberty to rename --runs by --batch_count, as this is what is displayed in the WebUi and mainly used by the stable diffusion community.

I also needed to tweak those argument default values in order to support this:

	--prompts: defaulted to ["cyberpunk forest by Salvador Dali"]
	--negative_prompts: defaulted to ["trees, green"]
	--seed: defaulted to -1
	--hf_model_id: defaulted to ""
	--write_metadata_to_png: defaulted to True

**Test session:**
I check a lot of combinations and didn't encountered problems so far, but it is pretty clear than covering all possible major cases is too much for me. If you think this is useful, it would be nice to have the community run further tests.

Thanks.

### DIFF
--- a/apps/stable_diffusion/src/utils/stable_args.py
+++ b/apps/stable_diffusion/src/utils/stable_args.py
@@ -17,15 +17,15 @@ p = argparse.ArgumentParser(
 p.add_argument(
     "-p",
     "--prompts",
-    action="append",
-    default=[],
+    nargs="+",
+    default=["cyberpunk forest by Salvador Dali"],
     help="text of which images to be generated.",
 )
 
 p.add_argument(
     "--negative_prompts",
     nargs="+",
-    default=[""],
+    default=["trees, green"],
     help="text you don't want to see in the generated image.",
 )
 
@@ -39,8 +39,8 @@ p.add_argument(
 p.add_argument(
     "--seed",
     type=int,
-    default=42,
-    help="the seed to use.",
+    default=-1,
+    help="the seed to use. -1 for a random one.",
 )
 
 p.add_argument(
@@ -48,7 +48,7 @@ p.add_argument(
     type=int,
     default=1,
     choices=range(1, 4),
-    help="the number of inferences to be made in a single `run`.",
+    help="the number of inferences to be made in a single `batch_count`.",
 )
 
 p.add_argument(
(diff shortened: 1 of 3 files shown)


apps/stable_diffusion/src/utils/stable_args.py
```diff
@@ -164,7 +164,7 @@ def path_expand(s):
 p.add_argument(
     "--hf_model_id",
     type=str,
-    default="stabilityai/stable-diffusion-2-1-base",
+    default="",
```
powderluv, line 167: Shouldn't we leave this as default ? For users that don't have any custom ckpt ?  And want a sane default ?

powderluv (approved):
Thank you for the clean up. Minor question on the hf_model_id below.

apps/stable_diffusion/src/utils/stable_args.py
drumicube: The current WebUi code is forcing it to "", so I don't think it will change anything
```
                    hf_model_id = gr.Textbox(
                        placeholder="Select 'None' in the Models dropdown on the left and enter model ID here e.g: SG161222/Realistic_Vision_V1.3",
                        value="",
                        label="HuggingFace Model ID",
                    )
```
But I just noticed the placeholder is refering to an unknown model to me, perhaps updating it with "stabilityai/stable-diffusion-2-1-base" would be preferable ?

drumicube: Yes, you're right !
I'm even having an issue using the command line script and connecting to hf, I didn't noticed it earlier.
I will rework that and test more.

PR #990, opened 2023-02-10
[SD][web] Add Img2Img UI

Shukla-Gaurav:


apps/stable_diffusion/web/ui/txt2img_ui.py
```diff
@@ -0,0 +1,246 @@
...
+from PIL import Image
+from apps.stable_diffusion.scripts import txt2img_inf
+from apps.stable_diffusion.src import (
+    prompt_examples,
+    args,
+    get_available_devices,
+)
+
+
+def resource_path(relative_path):
```
powderluv, line 15: We probably don't want to make copies of this

powderluv (approved):

apps/stable_diffusion/web/ui/txt2img_ui.py
Shukla-Gaurav: Yes. will take care of this in the next clean up patch.

PR #998, opened 2023-02-11
Launch dark theme by default

drumicube:
## Redirect to the dark theme at launch:
* I changed the approach: No css is used since it's not working with the new tabbed interface.
A small javascript is used to simply redirect to "?__theme=dark" if dark theme was configured (default)
* --theme=light option can be used to launch with the default white theme.
* I have removed my theme override from the css, but kept the original content. In this current state, I think it's safe to remove everything, as nothing is applied (footer visible)


powderluv:
Im fixing the CI with the python 3.11 move. Let me know if it is ready to land.

drumicube:
Yes, it's ok for me.
Good luck with the CI.

powderluv (commented):
All new files will have to be added to the shark_sd.spec file so it gets packaged into the .exe.  I will defer to @Shukla-Gaurav to review this change with JS etc. 

However if you can split the saving gradio image as a separate patch I can review / merge. In general good idea to keep one PR for one feature / bug fix.

drumicube:
To keep only one feature in this PR, I have removed the Gradio save tmp image to shark_tmp fix.
It now only contains one feature: Launch the dark theme by default
I will submit a new seperate PR for the shark_tmp fix later.

drumicube:
I'm closing this, as https://github.com/nod-ai/SHARK/pull/1007 should provide the same content.

PR #1006, opened 2023-02-13
Save gradio tmp images to shark_tmp folder and clean tmp images at launch

m68k-fr:
This PR moves tmp images generated by Gradio to the shark_tmp folder.
It also cleans all the gradio temp images from the last session at launch.

* An override of save_pil_to_file was used to redirect Gradio images output to ./shark_tmp/
* Clean the shark_tmp folder from all Gradio images (tmp*.png) at launch.
* To prevent a crash, it also recreate the shark_tmp folder if it was deleted by the end user.


Shukla-Gaurav (commented):
LGTM, please rename..
gradio_hacks.py -> web/utils/gradio_configs.py or gradio_utils.py something like that.

apps/stable_diffusion/web/gradio_hacks.py
```diff
@@ -0,0 +1,31 @@
+import os
+import tempfile
+import gradio
+from os import listdir
+
+gradio_tmp_imgs_folder = "./shark_tmp/"
```
Shukla-Gaurav, line 6: please update to abs path, use os.getcwd()

Shukla-Gaurav (changes requested):

apps/stable_diffusion/web/utils/gradio_configs.py
```diff
@@ -0,0 +1,31 @@
+import os
+import tempfile
+import gradio
+from os import listdir
+
+gradio_tmp_imgs_folder = os.getcwd() + "/shark_tmp/"
+
```
Shukla-Gaurav, line 7: Please fix lint error and also avoid using such additions. Use os.path.join() or Path() for concatenating directories.

Shukla-Gaurav (changes requested):

apps/stable_diffusion/web/utils/gradio_configs.py
m68k-fr: Hello, and thank for the feedback.
I'm not really sure for "shark_tmp/", I noticed there is also a os.sep, on the other hand, I didn't invented this path and found a "./shark_tmp/" in the existing code, so I supposed it's ok.
Sorry about that, I'm not very experienced in Python.

dan-garvey:
Unless @Shukla-Gaurav has any other concerns, we should get this in once ci passes. Thanks for your contribution!

Shukla-Gaurav (approved):
LGTM

PR #1022, opened 2023-02-15
add strength param

PhaneeshB:
- add strength param to img2img
use: `python apps/stable_diffusion/scripts/img2img.py --prompt=<prompt text> --img_path=<input image> --import_mlir --no-load_vmfb --hf_model="stabilityai/stable-diffusion-2-1-base" --clear_all --enable_stack_trace --scheduler=PNDM --strength=0.8`
- fix size issue

### DIFF
--- a/apps/stable_diffusion/scripts/img2img.py
+++ b/apps/stable_diffusion/scripts/img2img.py
@@ -77,7 +77,7 @@ def save_output_img(output_img):
         if args.write_metadata_to_png:
             pngInfo.add_text(
                 "parameters",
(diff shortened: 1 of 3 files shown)


apps/stable_diffusion/scripts/img2img.py
```diff
@@ -275,6 +279,15 @@ def img2img_inf(
...
     schedulers = get_schedulers(args.hf_model_id)
+    if args.scheduler != "PNDM":
+        if "Shark" in args.scheduler:
+            print(
+                f"SharkEulerDiscrete scheduler not supported. Switching to PNDM scheduler"
+            )
+            args.scheduler = "PNDM"
+        print(
+            f"[WARNING] Img2Img works best with PNDM scheduler. Please use that"
+        )
```
Shukla-Gaurav, line 290: put this in else block and return in this case.

Shukla-Gaurav (approved):
Thanks for the fix. LGTM!
Please add strength info in the `text_output` that we return in the web and print in the script.

PR #1037, opened 2023-02-16
temporarily xfail microsoft resnet50

dan-garvey:


powderluv (approved):

monorimet (approved):
I'll add Microsoft/resnet50 back in when it works. LGTM

PR #1039, opened 2023-02-16
[WebUi] Minor interface cleanup and Ui cosmetics

m68k-fr:
This PR contains minor WebUi cosmetic changes and few interface fixes:

* Interface displayed in full width for desktop devices.
* Gallery height overflow and gallery interface fixed for large batch count, batch_count=100 should be ok.
* White text logo from the header removed as this infos is now available in tab titles.
* All css static unit values replaced by the matching Gradio theme var.
* Minor css cosmetic changes. (margin, padding, text area, border radius, etc.)
* Pressing enter in the negative prompt box now trigger the pipeline (same as what is already available in the prompt box)

Responsive behaviour looked ok on all breakpoints.
Exe was built without noting an interface issue, but I was unable to render images as I also got the 'Cannot compile the model' error reported on Discord today. I didn't encountered this problem using the Python version.


Shukla-Gaurav (approved):
Thanks for these changes. Looks good to me.

PR #1040, opened 2023-02-16
img2img denoise strength

PhaneeshB:
- fix strength param to work with lower values of strength


PhaneeshB:
@powderluv  tested on cli and web; 
exe not tested (changes shouldn't affect .exe differently)

powderluv:
thanks will land once builder is green

powderluv (approved):

PR #1053, opened 2023-02-18
disable anythingv3 until issue is resolved

dan-garvey:


powderluv (dismissed):

build_tools/stable_diffusion_testing.py
```diff
@@ -54,6 +54,8 @@ def test_loop(device="vulkan", beta=False, extra_flags=[]):
         extra_flags.append("--beta_models=True")
     for import_opt in import_options:
         for model_name in hf_model_names:
+            if "Linaqruf" in model_name:
+                continue
```
powderluv, line 58: maybe good to qualify with the model and not everything from Linaqruf/ but it is ok to land for now.

powderluv (approved):

yzhang93:
What's the problem of anythingv3?

PR #1054, opened 2023-02-18
[WebUi] text-to-image: Update Gradio gallery on each inference

m68k-fr:
This PR refresh the text-to-image Gradio gallery on each inference, eliminating the need to wait for all batch_count to be processed before having images displayed in the WebUi:
The gallery will automatically refresh as soon as a new image is available.

As the gallery is fully refreshed by Gradio on each batch loop, this introduced 2 side effects:
* tmp_files: Gradio gallery beeing fully refreshed, all tmp files are recreated on each loop. This can lead to a huge number of tmp files generated for large batch count:
batch_loop=100 => (100x99x98...x2x1 tmp files, leading to about 5000 tmp images) I don't think this a a major issue for now, as the tmp folder is cleared at launch.
* Gallery log output: Similar problem, the txt2img_obj.log (Clip Inference time etc.) can be huge and intrusive. I simply decided to redirect the txt2img_obj.log output to the console instead of the WebUi output.


m68k-fr:
Put to draft: There's some side effects with Gradio beeing updated to V3.19.0

m68k-fr:
Sorry for the delay !

Indeed, a minor problem was introduced with the latest Gradio update and this "dynamic" gallery feature:
As, they have updated the progress bar (surely a higher z-index priority), interactions where not possible anymore with the gallery while it was building up, the progress bar beeing displayed in front, it captures all gallery mouse or keyboard inputs.
Simply launch the webUi with a --no-progress_bar argument and it solve the problem.
I also forced push a little update for a better renaming on the fly of  the "Generate image(s)" button.

Once again, sorry for the draft status and the annoyance...
LGTM this time :)

apps/stable_diffusion/scripts/txt2img.py
```diff
@@ -160,19 +167,32 @@ def txt2img_inf(
...
+        text_output += f"\nsize={height}x{width}, batch_count={batch_count}, batch_size={batch_size}, max_length={args.max_length}"
+        text_output += (
+            f"\n\nTotal image(s) generation time: {total_time:.4f}sec"
+        )
+        print(txt2img_obj.log)
+        txt2img_obj.log = ""
+        if i != batch_count - 1:
+            stable_diffusion_button_label = (
+                f"Processing batch {i+2} / {batch_count}"
+            )
```
Shukla-Gaurav, line 189: It would be better to disable the generate button instead of changing the labels.

apps/stable_diffusion/scripts/txt2img.py
```diff
@@ -160,19 +167,32 @@ def txt2img_inf(
...
+        else:
+            # This is the last inference: we restore the original button label
+            stable_diffusion_button_label = "Generate Image(s)"
+        yield generated_imgs, text_output, stable_diffusion_button_label
```
Shukla-Gaurav, line 194: we should only yield generated_imgs here. It's possible to yield subset of total number of outputs expected.

Shukla-Gaurav (changes requested):

apps/stable_diffusion/scripts/txt2img.py
m68k-fr: Labels are now kept unchanged.
A clear queue button was also added to cancel batches.
I gave up on tempering with buttons status, I spent more than 10 hours on it, it's a nightmare:
First it's not just buttons, prompt and neg prompt are also trigerring jobs, and what about fields from the other tabs ?
More over if we lock them, we will also need to unlock them.

I checked Auto1111, this is the direction they took. 
They added a state manager and update states on every possible actions, more over this is not working very well.
https://www.reddit.com/r/StableDiffusion/comments/y0mhkt/is_there_any_fix_for_the_gradio_bug_that_prevents/

apps/stable_diffusion/scripts/txt2img.py
m68k-fr: Output logs are now stricly identical to the original one, expect I report a progress in the output field when batch is running:
"Processing batch numberx / fromy"

m68k-fr:
A quick recap:
The feature itself is working, I'm using it for about 5 days now, it's really a game changer for me when producing images in large quantities, it adds a lot of comfort. No need to wait 15 minutes for an update or kill the app in case the prompt was misspelled.

Despite this, I highly recommend to not merge it:
Spamming generate & cancels buttons like crazy, I'm able to launch mutiple jobs simultaneously...
I could have missed something, but from what I found, it's seems there's not simple solution to fix this problem.
The problem also happends in Auto1111.

I will surely try to come back at this later.

PR #1057, opened 2023-02-19
Update setup_venv.ps1

cstueckrath:
make Python detection more robust.
Also support Python 3.11 if it is not the default Python version.


cstueckrath:
also your tests need an update

cstueckrath:
eh, I'll overhaul this venv script completely

PR #1059, opened 2023-02-19
[WebUi] css fix for Gradio v3.19.0

m68k-fr:
Gradio was updated to v3.19.0 few hours ago.
Here is the css fix.


powderluv (approved):

cstueckrath:
maybe set this together with a version pin in requirements.txt?

m68k-fr:
Today, they released V3.19.1 including an Ui fix.
I updated my venv, it looks good with this PR.
I also agree with @cstueckrath too, it would be nice to pin V3.19.* just to be safe.

powderluv (approved):

PR #1064, opened 2023-02-20
Update setup_venv.ps1

cstueckrath:
Make Python version detection more robust and support not default Python 3.11 installations

### DIFF
--- a/setup_venv.ps1
+++ b/setup_venv.ps1
@@ -26,12 +26,6 @@
 .OUTPUTS
   None
 
-.NOTES
-  Version        1.01
-  Author         powderluv, xzuyn, cstueckrath
-  Modification Date  2023-02-19
-  PurposeChange Find correct Python version
-
 #>
 
 param([string]$arguments)
--- a/setup_venv.ps1
+++ b/setup_venv.ps1
@@ -9,10 +9,10 @@
   If that environment does not exist, it creates it.
   
 .PARAMETER update-src
-  updates to latest version from git .\source
+  git pulls latest version
 
 .PARAMETER force
-  removes and recreates venv to force update all dependencies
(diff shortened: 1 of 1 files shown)


setup_venv.ps1
```diff
@@ -27,10 +27,10 @@
   None
 
 .NOTES
-  Version        1.0
-  Author         powderluv, xzuyn
-  Creation Date  2023-02-17
-  PurposeChange Initial script development
+  Version        1.01
```
powderluv, line 30: I'd like to avoid version numbers here and authors. Version should come from the CI and authors from git log

setup_venv.ps1
```diff
@@ -78,19 +66,27 @@ $version = if($p -is [System.Management.Automation.ErrorRecord])
...
-    $p
+    # otherwise return complete Python list
+    $PyVer = py --list
+}
+
+# deactivate any activated venvs
+if ($PyVer -like "*venv*")
+{
+  deactivate # make sure we don't update the wrong venv
+  $PyVer = py --list # update list
```
powderluv, line 77: Does this command work on all systems? I'm traveling and won't by and to check until Wednesday

powderluv (changes requested):

setup_venv.ps1
cstueckrath: this should work from at least Python 2.7
I didn't check further back.
I could not break it in Powershell 5 or Powershell 7 once any Version of Python is installed. And if that's not the case, you handle it further up.

setup_venv.ps1
cstueckrath: ok, will remove it from ps meta then.

powderluv (approved):

PR #1071, opened 2023-02-22
unpin diffuser to latest

powderluv:
Currently 0.13.x


powderluv:
@cstueckrath fyi. Back to tracking latest stable of diffusers now that the wget fix is in.

PhaneeshB (approved):

PR #1072, opened 2023-02-22
Add SD outpainting

jinchen62:
Add outpainting support on CLI
Flags:
--pixels: Number of pixels to be expanded (min=8, max=256, step=8)
--mask_blur: Number of blur pixels (min=0, max=64, step=1)
--left, --right, --top, --bottom: Expanding directions

`python apps/stable_diffusion/scripts/outpaint.py --prompt="your prompt" --img_path=path/to/img --import_mlir --hf_model_id="stabilityai/stable-diffusion-2-inpainting" --pixels=128 --mask_blur=8 --left --right --top --bottom --steps=20`

### DIFF
--- a/apps/stable_diffusion/scripts/outpaint.py
+++ b/apps/stable_diffusion/scripts/outpaint.py
@@ -1,0 +1,264 @@
+import sys
+import torch
+import time
+from PIL import Image
+from dataclasses import dataclass
+from apps.stable_diffusion.src import (
+    args,
+    OutpaintPipeline,
+    get_schedulers,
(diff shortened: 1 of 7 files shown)


requirements.txt
```diff
@@ -23,6 +23,7 @@ gradio
 altair
 omegaconf
 safetensors
+scikit-image
```
powderluv, line 26: do we need this ?

jinchen62: removed

powderluv (approved):

PR #1079, opened 2023-02-22
added an example to run sharded bloom

Eliasj42:
Updated sharding bloom to allow the compilation of huggingface bloom models into mlir


shark/examples/shark_inference/sharded_bloom.py
```diff
@@ -399,8 +674,10 @@ def download_560m(destination_folder):
 
     if args.device == "cuda" and args.device_list is not None:
         IS_CUDA = True
+        from cuda.cudart import cudaSetDevice
     if args.download:
-        download_560m(args.model_path)
+        # download_model(args.model_path, args.model_name)
```
powderluv, line 679: shouldn't the download / create be two options ?

powderluv (approved):

PR #1080, opened 2023-02-22
Recursively import transformers to avoid __init__.py issues

powderluv:
Also drop torchvision. The test passed and didn't fail but we can't be sure it fixes the __init__.py issue yet.


powderluv:
@monorimet fyi dropping torchvision from the windows build. I think the Linux one also has it dropped unless you use IMPORTER=1 (and I think we can drop it from there too ?)

monorimet (approved):
LGTM. Windows builder doesn't rely on torchvision for pytests but it is needed for generate_sharktank (so we should keep it as a Linux importer req)

PR #1082, opened 2023-02-23
Resize input image and mask for SD inpainting

jinchen62:


powderluv (approved):
lets refine more after this lands

PR #1086, opened 2023-02-23
Disable tuned configs on all inpainting models

yzhang93:


dan-garvey (approved):

apps/stable_diffusion/src/utils/utils.py
```diff
@@ -244,11 +244,7 @@ def set_init_device_flags():
         base_model_id = fetch_and_update_base_model_id(args.ckpt_loc)
 
     if (
-        args.hf_model_id
-        in [
-            "runwayml/stable-diffusion-inpainting",
-            "stabilityai/stable-diffusion-2-inpainting",
-        ]
+        "inpainting" in args.hf_model_id
```
powderluv, line 247: It's ok for now but I think we can't assume the model name has inpainting in it

powderluv (approved):

PR #1089, opened 2023-02-24
Fix outpainting params

jinchen62:
- fix output width and height
- fix input params pixels and mask_blur


powderluv (changes requested):
I think people are using custom sizes too. Not just 512^2

powderluv (approved):

PR #1098, opened 2023-02-25
[Fix] Unsupported width and height argument error

m68k-fr:
Hello,

Since https://github.com/nod-ai/SHARK/pull/1089 was merged on main, I got the following error using --width=768
```
index.py: error: argument --width: invalid choice: 768 (choose from 384, 392, 400, 408, 416, 424, 432, 440, 448, 456, 464, 472, 480, 488, 496, 504, 512, 520, 528, 536, 544, 552, 560, 568, 576, 584, 592, 600, 608, 616, 624, 632, 640, 648, 656, 664, 672, 680, 688, 696, 704, 712, 720, 728, 736, 744, 752, 760)
```
* The range was set to 768 max, but it looks like the upper limit is not included in the accepted  values. Same for height.
I simply increment by 1 all those ranges and updated the corresponding inputs in the WebUi to match those new limits.
* Placeholder exemples for _HuggingFace Model ID_  in the inpainting and outpainting tabs was also replaced by an inpainting model Id as those fields only accept that.

Please, let me know if I missed something.
Once again, thanks a lot for this cool project and all the hard work.
👍


Shukla-Gaurav (approved):
LGTM! Thanks for the contribution. CI failure looks unrelated.

m68k-fr:
You're welcome.
:)

PR #1103, opened 2023-02-28
Fix loading custom inpainting models

### LINKED ISSUE #1088: Build 557 custom inpainting models
Through a lot of testing today of inpainting on a 7900xtx with build 557

I found that custom inpainting models currently only work when compiling from huggingface ID

Example model: ghunkins/stable-diffusion-liberty-inpainting

1. Put ghunkins/stable-diffusion-liberty-inpainting into the HuggingFace Model ID field
2. use a 512x512 input image and generate at default settings 512x512
3. profit

However if downloading the same inpainting model or any other custom model it doesn't work currently.

1. Download https://huggingface.co/ghunkins/stable-diffusion-liberty-inpainting/blob/main/liberty_main-inpainting.safetensors
2. Place it into ./models directory
4. Launch shark and select the model from drop down
5. use a 512x512 img and try generating at default settings 512x512
6. produces errors [safetensorsmodel error.txt](https://github.com/nod-ai/SHARK/files/10819746/safetensorsmodel.error.txt)

Now I have successfully got custom inpainting .safetensors models I've made in automatic1111 as well as downloaded models like the liberty-inpainting model and others working but only by using Onnx to extract the diffuser files, then copying those diffuser files into the ./models/(model_name)/ folder and having shark compile the model with the .safetensors file  and the diffusers onnx extracted. but that is obviously a bit too complicated.

I've also found that input image must match dimensions of the selection in the GUI. For example 512x512 default resolutions must also have input image of 512x512 as well. This is probably a separate "issue" but wanted to put this here to avoid any additional confusion/errors when testing.

jinchen62:
fix #1088


apps/stable_diffusion/src/models/opt_params.py
```diff
@@ -16,8 +16,6 @@
     "stabilityai/stable-diffusion-2-1": ["stablediffusion", "v2_1base"],
     "stabilityai/stable-diffusion-2-1-base": ["stablediffusion", "v2_1base"],
     "CompVis/stable-diffusion-v1-4": ["stablediffusion", "v1_4"],
-    "runwayml/stable-diffusion-inpainting": ["stablediffusion", "inpaint_v1"],
-    "stabilityai/stable-diffusion-2-inpainting": ["stablediffusion", "inpaint_v2"],
```
yzhang93, line 20: You should probably keep these two lines, otherwise there will be problem loading tuned configs for inpainting,

powderluv (dismissed):

apps/stable_diffusion/src/utils/resources/base_model.json
```diff
@@ -111,7 +111,7 @@
             }
         }
     },
-    "runwayml/stable-diffusion-inpainting": {
+    "stabilityai/stable-diffusion-2-inpainting": {
```
yzhang93, line 114: Is there a reason why you change the order of the two models?

yzhang93 (approved):

PR #1105, opened 2023-02-28
Fix call to Torch Inductor

mariecwhite:
Updates call to enable Torch Inductor.

To get Torch Inductor working, first update the version of Torch being run:

```
pip install numpy --pre torch[dynamo] --force-reinstall --extra-index-url https://download.pytorch.org/whl/nightly/cu117
```

This may cause dependency conflicts with `torch-mlir` so Torch imports may break.

Then uncomment the line:

```
#frontend_model = torch.compile(frontend_model, mode="max-autotune", backend="inductor")
```

in shark/shark_benchmark_runner.py

### DIFF
--- a/shark/shark_benchmark_runner.py
+++ b/shark/shark_benchmark_runner.py
@@ -118,9 +118,10 @@ class SharkBenchmarkRunner(SharkRunner):
         )
         HFmodel, input = get_torch_model(modelname)[:2]
         frontend_model = HFmodel.model
-        # frontend_model = dynamo.optimize("inductor")(frontend_model)
         frontend_model.to(torch_device)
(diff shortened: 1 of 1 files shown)


powderluv:
We can update upstream torch-mlir to install with [dynamo]

powderluv (approved):

mariecwhite:
> We can update upstream torch-mlir to install with [dynamo]

That would be great.

msaroufim:
Just FYI you can run the installation command without `[dynamo]` now and it should work

```
pip install numpy --pre torch --force-reinstall --extra-index-url https://download.pytorch.org/whl/nightly/cu117

```

PR #1110, opened 2023-02-28
[SD] Stencil execution

Abhishek-Varma:


apps/stable_diffusion/src/utils/stencils/stencil_utils.py
```diff
@@ -63,7 +63,7 @@ def controlnet_hint_shaping(
             return controlnet_hint
         else:
             raise ValueError(
-                f"Acceptble shape of `controlnet_hint` are any of ({channels}, {height}, {width}),"
+                f"Acceptble shape of `stencil_hint` are any of ({channels}, {height}, {width}),"
```
PhaneeshB, line 66: nit : just stencil would be good too

PR #1120, opened 2023-03-01
Run windows pytests in subprocesses

monorimet:
Also removes a command in the workflow .yaml that prints bench_results.csv (which won't be generated since benchmarks are disabled)


monorimet:
pytest-forked does not support Windows -- other options for managing memory effectively in pytest must be explored. Closing this PR for now.

PR #1126, opened 2023-03-02
Use general way to identify custom inpainting models

jinchen62:


apps/stable_diffusion/scripts/outpaint.py
```diff
@@ -200,14 +200,16 @@ def outpaint_inf(
...
-        print("Please use inpainting model with --hf_model_id.")
-        exit()
 
     dtype = torch.float32 if args.precision == "fp32" else torch.half
     cpu_scheduling = not args.scheduler.startswith("Shark")
     set_init_device_flags()
-    schedulers = get_schedulers(args.hf_model_id)
+    model_id = (
+        args.hf_model_id
+        if "inpaint" in args.hf_model_id
```
dan-garvey, line 209: Should this be outpaint?

dan-garvey (approved):
As long as it passes SD tests fine to merge after addressing comment

apps/stable_diffusion/scripts/outpaint.py
jinchen62: Outpainting also uses inpainting models.

PR #1128, opened 2023-03-02
In models testing fix paths to be relative to the script dir not cwd

sogartar:


monorimet (approved):
Thanks for this! LGTM

PR #1129, opened 2023-03-02
In SD don't include device path in vmfb filename

sogartar:
Include only the driver name instead.

### DIFF
--- a/apps/stable_diffusion/src/utils/utils.py
+++ b/apps/stable_diffusion/src/utils/utils.py
@@ -25,11 +25,7 @@ from diffusers.pipelines.stable_diffusion.convert_from_ckpt import (
 
 
 def get_extended_name(model_name):
-    device = (
-        args.device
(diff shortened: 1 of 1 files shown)


sogartar:
At some point it will be probably better to have the distinction between target backend and driver. The same vmfb compiled for one target backend may be used by different drivers.
If we are somehow querying the device for its capabilities and then compiling based on that then it makes sense to include the device id in the path.

powderluv (commented):
wont this clash if we tried to build for cuda and vulkan on the same device like A100 ?

sogartar:
> wont this clash if we tried to build for cuda and vulkan on the same device like A100 ?

Logically they will be different devices. They would be named `cuda://...` and `vulkan://...` respectively. The file names would then contain just `cuda` and `vulkan`.

powderluv (approved):

PR #1139, opened 2023-03-03
Install IREE pre-releases

mariecwhite:
The latest version of IREE being installed in SHARK is 20230221.437 (see [summary.html](https://storage.googleapis.com/shark-benchmark-artifacts/latest/summary.html)). Update `setup_venv.sh` to also install pre-releases.


monorimet (approved):
Thanks! LGTM

PR #1140, opened 2023-03-03
Add Stable diffusion fine tuning script

vivekkhandelwal1:
This commit adds the sd fine tuning script which runs through the torchdynamo path.

### DIFF
--- a/shark/examples/shark_training/stable_diffusion/stable_diffusion_fine_tuning.py
+++ b/shark/examples/shark_training/stable_diffusion/stable_diffusion_fine_tuning.py
@@ -1,0 +1,899 @@
+# Install the required libs
+# pip install -U git+https://github.com/huggingface/diffusers.git
+# pip install accelerate transformers ftfy
+
+# Import required libraries
+import argparse
+import itertools
+import math
+import os
+from typing import List
+import random
+
+import numpy as np
+import torch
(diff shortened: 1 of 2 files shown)


shark/examples/shark_training/stable_diffusion/stable_diffusion_fine_tuning.py
```diff
@@ -0,0 +1,899 @@
...
+while not os.path.exists(str(out_dir)):
+    try:
+        os.mkdir(out_dir)
+    except OSError as error:
+        print("Output directory not created")
+
+###### Torch-MLIR Compilation ######
+
+
+def _remove_nones(fx_g: torch.fx.GraphModule) -> List[int]:
```
pashu123, line 532: Please move all the shark-related stuff to another file. This can be used to train or run other models as well. We should also think about adding this in the core-api.

pashu123: torchdynamo and torch-mlir compilation too.

vivekkhandelwal1: I will do this in another PR. Can we please get this merged if there is/are no major issue(s)?

pashu123 (approved):

PR #1147, opened 2023-03-05
[WebUi] txt2img_ui: Import png metadata

m68k-fr:
Load and apply png metadata using an Image field in the txt2img tab

As long as the source PNG was generated by Shark, it should be ok for all fields (local ckpt/safetensors, predefined models like AnythingV3 or hf model ids are supported)
Drag&dropping images from civitai also worked, but minor fixes need to be done to the png info itself, I will focus on that later.

To avoid a tedious review, it's also limited to the main feature itself for now:
* No fancy drag&drop in the prompt textfield. As far as I know, this is not supported natively by Gradio, a javascript hack is required. So it's just a simple image field for now.
* Only available in txt2img tab for now.

### DIFF
--- a/apps/stable_diffusion/web/utils/png_metadata.py
+++ b/apps/stable_diffusion/web/utils/png_metadata.py
@@ -1,0 +1,121 @@
+import re
+import os
+from pathlib import Path
+from apps.stable_diffusion.web.ui.txt2img_ui import (
+    png_info_img,
+    prompt,
+    negative_prompt,
+    steps,
+    scheduler,
+    guidance_scale,
+    seed,
+    width,
+    height,
+    custom_model,
+    hf_model_id,
+)
+from apps.stable_diffusion.web.ui.utils import (
+    get_custom_model_path,
+    scheduler_list_txt2img,
+    predefined_models,
+)
+
+re_param_code = r'\s*([\w ]+):\s*("(?:\\"[^,]|\\"|\\|[^\"])+"|[^,]*)(?:,|$)'
+re_param = re.compile(re_param_code)
+re_imagesize = re.compile(r"^(\d+)x(\d+)$")
+
+
+def parse_generation_parameters(x: str):
+    res = {}
+    prompt = ""
+    negative_prompt = ""
+    done_with_prompt = False
+
+    *lines, lastline = x.strip().split("\n")
+    if len(re_param.findall(lastline)) < 3:
+        lines.append(lastline)
+        lastline = ""
+
+    for i, line in enumerate(lines):
+        line = line.strip()
+        if line.startswith("Negative prompt:"):
+            done_with_prompt = True
+            line = line[16:].strip()
+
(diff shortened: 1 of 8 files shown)


apps/stable_diffusion/web/utils/png_metadata.py
```diff
@@ -0,0 +1,138 @@
...
+    prompt,
+    negative_prompt,
+    steps,
+    sampler,
+    cfg_scale,
+    seed,
+    width,
+    height,
+    custom_model,
+    hf_model_id,
```
Shukla-Gaurav, line 77: The input should be only pil_data, isn't it?
You can always return selected fields(check return by dict gradio) in order to retain the field values which are not updated here.

Shukla-Gaurav (changes requested):
LGTM! Thank you for contribution. Just one comment to address.

apps/stable_diffusion/web/utils/png_metadata.py
m68k-fr: Hello.
Thanks for the feedback, I saw the return by dict gradio, but didn't pay a lot of attention at that time, that was a great mistake, this stuff is super useful, thanks :)
So, I tested it using dict, and indeed it's far better.
```
        return {
            png_info_img: None,
            prompt: pil_prompt,
            negative_prompt: metadata["Negative prompt"],
            steps: int(metadata["Steps"]),
            scheduler: pil_scheduler,
            guidance_scale: float(metadata["CFG scale"]),
            seed: int(metadata["Seed"]),
            width: float(metadata["Size-1"]),
            height: float(metadata["Size-2"]),
            custom_model: png_custom_model,
            hf_model_id: png_hf_model_id,
        }
```

But I have two concerns about this method:

1 - Very minor: As the import_png_metadata function is located outside of the txt2img_ui file, I need to import all those ui blocks in the png_metadata file, but this is causing a circular dependencies problem. I managed to fix it by moving the import after all the ui blocks were declared. It works but that's not super elegent:
```
        from apps.stable_diffusion.web.utils.png_metadata import import_png_metadata
        png_info_img.change(
            fn=import_png_metadata,
```

2 - More concerning: If I'm not wrong, import_png_metadata is now tied to the txt2img_ui tab. I think I won't be able to use it on the other tabs. And this was my next goal.

If I missed something, or have a solution for this problem, please let me know.

As a side note, I will surely post a second PR this week to expend the import png to the other tabs.
Perhaps it would be better to report this problem to the next PR: It is pretty obvious that I will have to refactor the png import ifself to handle multiple tabs. I will keep in mind the dict method and see if I can make it work with multiples sources.

Thanks

m68k-fr: The requested changes had been made.

Shukla-Gaurav: Thanks

PR #1165, opened 2023-03-10
added more memory efficient method to run large bloom models with sha…

Eliasj42:
…rded blooms

I recreated this PR with one squashed commit


powderluv (changes requested):
Can we just have it in one file and add a flag if we need some special flow?

PR #1166, opened 2023-03-10
[SD] Use dynamic stencil HF repo id

Abhishek-Varma:
-- This commit removes the hardcoded HF ID for Stencil and instead
   utilizes a dynamic instantiation of HF model.


PhaneeshB (approved):
LGTM

PR #1175, opened 2023-03-12
[Web] Improve dropdowns ux

m68k-fr:
Hello, 
This should fix most of the Ui display problems introduced by Gradio 3.20

* Hide the download icon from the nod logo.
* Hide delete buttons visible in dropdowns.
* Hide selected icons visible next to dropdown entries.
* Display model names only (without fullpath) in model dropdowns.
* ckpt & safetensors names alphabetically sorted in model dropdowns.
* Renamed the "Strength" label in image-to-image tab by "Denoising Strength".

Note: For the download logo icon, I took the fast and lazy path, just hiding it. A safer option would have been to replace the image component by a html one. Perhaps for a later upgrade if a logo problem pop again.

Tested on all tabs, python version, windows binary distribution, and txt2img command line.

Have a nice week, and thanks a lot for this really cool project. :)

### DIFF
--- a/apps/stable_diffusion/scripts/img2img.py
+++ b/apps/stable_diffusion/scripts/img2img.py
@@ -3,6 +3,7 @@ import torch
 import time
 from PIL import Image
 from dataclasses import dataclass
+from apps.stable_diffusion.web.ui.utils import get_custom_model_pathfile
 from apps.stable_diffusion.src import (
     args,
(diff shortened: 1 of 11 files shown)


powderluv (approved):

powderluv:
Lets land this so it doesn't as bad. Once @Shukla-Gaurav is on we can address any comments he may have.

PR #1182, opened 2023-03-13
Add flag to augment the device allocator

sogartar:
Example:
$ python my_app.py --device_allocator caching debug
This will wrap the device allocator with first caching allocator then debug allocator.

$ python my_app.py --device_allocator caching
Only wrap with caching allocator.


powderluv (approved):

powderluv:
@Shukla-Gaurav can you confirm this works through to the web UI too

PR #1194, opened 2023-03-15
Clear memory cache when switching model and mode

jinchen62:
Using the global objects instead of creating one for each mode, so it won't create extra pipeline object when switching modes. Also clearing the cache when switching models to avoid memory leak.

### DIFF
--- a/apps/stable_diffusion/scripts/inpaint.py
+++ b/apps/stable_diffusion/scripts/inpaint.py
@@ -1,7 +1,6 @@
 import torch
 import time
 from PIL import Image
-from dataclasses import dataclass
 from apps.stable_diffusion.src import (
     args,
     InpaintPipeline,
@@ -13,20 +12,6 @@ from apps.stable_diffusion.src import (
 )
 
 
-@dataclass
-class Config:
-    model_id: str
-    ckpt_loc: str
-    precision: str
-    batch_size: int
-    max_length: int
-    height: int
-    width: int
-    device: str
-
-
-inpaint_obj = None
-config_obj = None
 schedulers = None
 
 # set initial values of iree_vulkan_target_triple, use_tuned and import_mlir.
@@ -58,10 +43,12 @@ def inpaint_inf(
     save_metadata_to_json: bool,
     save_metadata_to_png: bool,
 ):
-    from apps.stable_diffusion.web.ui.utils import get_custom_model_pathfile
+    from apps.stable_diffusion.web.ui.utils import (
+        get_custom_model_pathfile,
+        Config,
+    )
+    import apps.stable_diffusion.web.utils.global_obj as global_obj
 
-    global inpaint_obj
-    global config_obj
     global schedulers
 
     args.prompts = [prompt]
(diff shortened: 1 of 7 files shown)


apps/stable_diffusion/scripts/txt2img.py
```diff
@@ -103,6 +89,7 @@ def txt2img_inf(
     dtype = torch.float32 if precision == "fp32" else torch.half
     cpu_scheduling = not scheduler.startswith("Shark")
     new_config_obj = Config(
+        "txt2txt",
```
powderluv, line 92: is this correct ?

powderluv (commented):
adding a new .py file usually has some issue with .exe. So please build / test the .exe and run it. 

(tip you have to run the process_skipfiles.py before calling pyinstaller. @dan-garvey can help if you want. 

Also please test the new  stencils are still working with this change.

apps/stable_diffusion/web/utils/global_obj.py
```diff
@@ -0,0 +1,41 @@
+import gc
+import torch
```
powderluv, line 2: wait what !!! Is torch holding up the memory ?

apps/stable_diffusion/web/utils/global_obj.py
```diff
@@ -0,0 +1,41 @@
...
+    return config_obj
+
+
+def clear_cache():
+    global sd_obj
+    global config_obj
+    del sd_obj
+    del config_obj
+    gc.collect()
+    torch.cuda.empty_cache()
```
powderluv, line 41: Do we need torch.cuda ? What happens when there is no cuda

jinchen62: I had attempts with separate sd pipeline objects and I remembered I need this probably because I'm with nvidia gpu, but we shouldn't need it now since I just maintain one global pipeline object. Tested and removed.

apps/stable_diffusion/scripts/inpaint.py
```diff
@@ -97,6 +84,7 @@ def inpaint_inf(
     dtype = torch.float32 if precision == "fp32" else torch.half
     cpu_scheduling = not scheduler.startswith("Shark")
     new_config_obj = Config(
+        "inpaint",
```
powderluv, line 87: remove ?

powderluv: oh never mind. I think we need this

powderluv (approved):

dan-garvey:
@powderluv jinchen doesnt have access to a machine so if we can rebase this on top of my exe test patch the ci can test it for him

apps/stable_diffusion/scripts/inpaint.py
```diff
@@ -105,9 +93,15 @@ def inpaint_inf(
         height,
         width,
         device,
+        None,
```
dan-garvey, line 96: for unclear args please add the argument_name=None

apps/stable_diffusion/scripts/outpaint.py
```diff
@@ -107,9 +95,15 @@ def outpaint_inf(
         height,
         width,
         device,
+        None,
+        None,
```
dan-garvey, line 99: same as above

dan-garvey (changes requested):
cool patch. some comments to explain what you told me over chat would be helpful for maint. especially in global_obj.py

powderluv:
is this good to land ?

powderluv (approved):

PR #1211, opened 2023-03-19
add pytorch_lightning to requirements

cstueckrath:
this will additionally add lightning-utilities and torchmetrics


powderluv (commented):
We will also need to add it to the .spec files in apps/stable_diffusion/ so it shows up for exe users.

cstueckrath:
hope this works that way? Haven't touched anything like this before.

powderluv (approved):

powderluv:
thank you. that _should_ work. We can test and iterate.

PR #1212, opened 2023-03-19
[Web] Stop images

m68k-fr:
"Stop batch" will now instantly cancel images generation for txt2img, img2img, inpaint and outpaint.
Sorry about the long PR, to avoid adding more, I have reported the implementation for the LoRA training and the upscaler tabs.

**Principle:** 
A pipeline status variable had been added with only two possible state for now (IDLE, CANCEL)
When the UI send a cancel signal, the pipeline stop iterating and return immediately.
The canceled image is skipped (not saved on disk and not displayed in the ui) as it is incomplete.

This is just an entry step, more work will be needed to finish a fully working "stop" feature.
If this is accepted, here is a list of possible updates I would like to add later:

	* Implement Upscaler and LoRa training tabs.
	* Manage more pipeline states: EXPORT_DIFFUSER, COMPILE_MODEL, LOAD_MODEL (set on first model load or when switching models), GENERATE_IMAGES.
	* Add a status message textfield in the UI header, and refresh it with a user friendly message depending on the current pipeline status.
	* Report the currently processed batch in the message textfield (Generating image x from y)
	* As instant cancel is not possible or recommended on some state (EXPORT_DIFFUSER, COMPILE_MODEL, LOAD_MODEL): Report a "Canceling please wait" in the status message textfield.
	* Move back the "Stop batch" button and rename it to "Stop" when all it ok.
	
**Identified bug:** 
Spamming at a crazy pace Generate and Stop buttons can report an error, or even save the last partially generated image on disk. This is not fatal and the app will continue to work, but further debugging will be needed to fix it.

### DIFF
--- a/apps/stable_diffusion/scripts/img2img.py
+++ b/apps/stable_diffusion/scripts/img2img.py
@@ -12,6 +12,7 @@ from apps.stable_diffusion.src import (
     clear_all,
     save_output_img,
 )
+from apps.stable_diffusion.src.utils import get_generation_text_info
 
 
 schedulers = None
@@ -79,6 +80,9 @@ def img2img_inf(
         Config,
     )
     import apps.stable_diffusion.web.utils.global_obj as global_obj
+    from apps.stable_diffusion.src.pipelines.pipeline_shark_stable_diffusion_utils import (
+        SD_STATE_CANCEL,
+    )
 
     global schedulers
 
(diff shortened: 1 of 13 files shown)


powderluv (approved):

powderluv:
Thank you. I like this since it builds on GC and doesn't need to plumb into the low level interfaces. Would be good to keep testing to make sure the clean up happened correctly

m68k-fr:
Same, that's really cool to avoid touching the low level , it's all thanks to the global_obj and GC introduced by @jinchen62, mega kudos to him 👍 
I agree too with testing more before landing, clearly more than one commit will be needed to finish this.
I will continue next week to tackle the annoying bug.
See you.

powderluv:
Do you want to do more testing or is it good enough to land ?

m68k-fr:
Sorry, it should be ok for this one.

PR #1219, opened 2023-03-20
Fix t5 models

mariecwhite:
Some changes that got left out during the merge.


monorimet (approved):
Thanks for the follow up! LGTM

PR #1235, opened 2023-03-22
Remove deprecated cli option iree-hal-cuda-disable-loop-nounroll-wa

ThomasRaoux:


ThomasRaoux:
Do you know if there are other repos than need the same change?

monorimet:
> Do you know if there are other repos than need the same change?

Yes, it seems it still exists in transformer-benchmarks. I will double-check the SHARK repo for any other instances of this flag.

monorimet:
It appears that all instances of this flag in SHARK are covered by your commit.

PR #1236, opened 2023-03-22
[Web] Add state manager and status bar

m68k-fr:
Rework of the stop batch button for a more global app state approach:

* Revert the code added to global_obj to handle the pipeline status.
* A new minimal and dedicated State class was added with a app singleton to manage app states.
* A status bar is displayed on top of each Ui tabs and refreshed every 2 seconds to display the current app state.
* A basic list of states is managed for now: Ready / Loading Models x / Running Job: x% / Canceling,please wait


m68k-fr:
I have also prepatched the upscaler tab, but I wasn't able to fully test it:
```
RuntimeError: The size of tensor a (128) must match the size of tensor b (256) at non-singleton dimension 1
```
The error is catched, and will put back the Ui in ready state to prevent a locked interface.

As this PR is quite long, I haven't done a thing for the upcoming loRA training tab.
I advice to pay a special attention to states before reactivating it.

m68k-fr:
Put to draft as a major refactoring will be needed once the Ui update is done.

PR #1242, opened 2023-03-24
Load and unload SD models on demand

jinchen62:
#1203 
- check "Low VRAM" checkbox on webui or use --ondemand flag on cli
- TODO: unload clip

### DIFF
--- a/apps/stable_diffusion/src/models/model_wrappers.py
+++ b/apps/stable_diffusion/src/models/model_wrappers.py
@@ -11,7 +11,7 @@ from apps.stable_diffusion.src.utils import (
     get_opt_flags,
     base_models,
     args,
-    fetch_vmfbs,
+    fetch_vmfb,
     preprocessCKPT,
     get_path_to_diffusers_checkpoint,
     fetch_and_update_base_model_id,
@@ -53,6 +53,11 @@ def replace_shape_str(shape, max_len, width, height, batch_size):
         else:
             new_shape.append(shape[i])
     return new_shape
+
+
+def check_compilation(model, model_name):
+    if not model:
+        raise Exception(f"Could not compile {model_name}. Please create an issue with the detailed log at https://github.com/nod-ai/SHARK/issues")
 
 
 class SharkifyStableDiffusionModel:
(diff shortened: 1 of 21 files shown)


powderluv:
is this ok to update and land

apps/stable_diffusion/src/models/model_wrappers.py
```diff
@@ -549,138 +565,161 @@ def process_custom_vae(self):
...
+        if self.base_model_fetched != "":
+            return self.compile_model(self.base_model_fetched, "clip")
+        for model_id in base_models:
+            try:
+                compiled_clip = self.compile_model(model_id, "clip")
+            except Exception as e:
+                print(e)
+                print("Retrying with a different base model configuration")
+                continue
+            fetch_and_update_base_model_id(self.model_to_run, model_id)
```
Abhishek-Varma, line 620: I see instances like these all over.
This is completely wrong for 2 reasons :-
1. You're performing the "retry" for each individual models. So, theoretically, if there were 'N' retries previously - this would now involve a total of ~3N-4N retries.
2. For each model you're invoking `fetch_and_update_base_model_id`! This is so wrong which is apparent in the following scenario :-
 Say, UNET for `v1.5` compiles first then we'd map `model_to_run` with "CompVis v1.4". After this when VAE compiles it'd end up replacing "v2.1" with, say, "upscaler" base ID. And that's wrong because now we're in the state where "v1.5" is being mapped to "v2.1", instead of "v1.4" in variants.json, and therefore in the next invocation, if someone uses `--clear_all`, because of https://github.com/nod-ai/SHARK/blob/e9f45d6b9d1712f2a5d3bb3944f6b1dc5536f967/apps/stable_diffusion/src/models/model_wrappers.py#L166 the system would crash because you're trying to compile "v1.5" Unet with "v2.1"'s input configuration due to : https://github.com/nod-ai/SHARK/blob/e9f45d6b9d1712f2a5d3bb3944f6b1dc5536f967/apps/stable_diffusion/src/models/model_wrappers.py#L633-L634

Abhishek-Varma (changes requested):

Abhishek-Varma:
Also, can you rebase/adjust your changes on top of https://github.com/nod-ai/SHARK/pull/1248 (I'd suggest you use a separate branch for this) and check if the delta of things you've solved in this PR still holds good?

In that PR I've taken care of individual control of model's compilation within one function itself. Guess you'd need to simply create individual functions for each now, similar to the way you were trying in this PR, just leaner minus the retries.

PR #1246, opened 2023-03-26
fix Python version detection bug

cstueckrath:


setup_venv.ps1
```diff
@@ -78,12 +78,12 @@ if (!($PyVer.length -ne 0)) {$p} # return Python --version String if py.exe is u
 if (!($PyVer -like "*3.11*") -and !($p -like "*3.11*")) # if 3.11 is not in any list
 {
     Write-Host "Please install Python 3.11 and try again"
-    break
+    exit 34
```
powderluv, line 81: what is exit 34 ?

cstueckrath: a different exit code, here requirements_not_met
Doesn't really matter for running locally, but maybe interesting if automated -> horde, parent applications, rich-UIs...

powderluv: Would be nice to have a sequential increase in the number.

powderluv (approved):

setup_venv.ps1
cstueckrath: I don't know how you want to handle this, but I'd group return codes based on reason like HTTP codes do.
There is no reason for them to be sequential, see https://learn.microsoft.com/en-us/windows/win32/msi/error-codes
Until you have documentation we can just keep it as-is, but I would avoid using 1-digit codes except 0 (no error) and 1 (general failure) as those are very low-level and often system reserved.

powderluv (approved):

PR #1257, opened 2023-03-28
allow pytest to retry getting model artifacts + disable autotuning for pytorch benchmarks

monorimet:


monorimet:
FYI @mariecwhite to unblock SHARK/IREE benchmarks on GPU + revive our CI while I work on the benchmark framework

shark/shark_downloader.py
```diff
@@ -204,17 +204,17 @@ def download_model(
...
-        tank_dir = WORKDIR
-        gen_shark_files(model_name, frontend, tank_dir, import_args)
-        with open(filename, mode="rb") as f:
-            mlir_file = f.read()
+        print(
+            "The model data was not found. Trying to generate artifacts locally."
+        )
+        gen_shark_files(model_name, frontend, WORKDIR, import_args)
 
+    assert os.path.exists(filename)
```
dan-garvey, line 215: should provide error message if fail on assert due to filename still not existing after gen

tank/test_models.py
```diff
@@ -169,18 +169,27 @@ def create_and_check_module(self, dynamic, device):
...
-                tank_url=self.tank_url,
-                frontend=self.config["framework"],
-                import_args=import_config,
-            )
-        except NoImportException:
-            pytest.xfail(
-                reason=f"Artifacts for this model/config must be generated locally. Please make sure {self.config['framework']} is installed."
-            )
-
+        dl_gen_attempts = 3
```
dan-garvey, line 172: If we need this to be greater than 2 I feel like we have bigger problems. Are we hitting spurious failures often enough that a single retry wouldn't cover them?

shark/shark_benchmark_runner.py
```diff
@@ -132,12 +132,12 @@ def benchmark_torch(self, modelname):
         frontend_model.to(torch_device)
         input.to(torch_device)
 
-        try:
-            frontend_model = torch.compile(
-                frontend_model, mode="max-autotune", backend="inductor"
-            )
-        except RuntimeError:
-            frontend_model = HFmodel.model
+        # try:
```
dan-garvey, line 135: dead code?

dan-garvey (changes requested):

shark/shark_benchmark_runner.py
monorimet: Will need to re-enable as soon as CUDA context issues from running inductor in a forked subprocess are resolved.

tank/test_models.py
monorimet: It doesn't need to be greater than 2. I'll reduce it

dan-garvey (dismissed):

shark/shark_benchmark_runner.py
dan-garvey: Ideally add a comment to that effect

dan-garvey (approved):

PR #1260, opened 2023-03-29
[SD] Fix custom model path for WebUI

Abhishek-Varma:
-- This commit fixes custom model path for WebUI.

### DIFF
--- a/apps/stable_diffusion/web/ui/utils.py
+++ b/apps/stable_diffusion/web/ui/utils.py
@@ -75,14 +75,20 @@ def resource_path(relative_path):
 
 
 def get_custom_model_path(model="models"):
-    prefix = Path(args.ckpt_dir) if args.ckpt_dir else Path(Path.cwd())
+    # If `--ckpt_dir` is provided it'd override the heirarchical folder
+    # structure in WebUI :-
+    #       model
+    #         |___lora
+    #         |___vae
+    if args.ckpt_dir:
+        return Path(args.ckpt_dir)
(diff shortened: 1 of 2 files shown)


apps/stable_diffusion/web/ui/utils.py
```diff
@@ -75,14 +75,13 @@ def resource_path(relative_path):
 
 
 def get_custom_model_path(model="models"):
-    prefix = Path(args.ckpt_dir) if args.ckpt_dir else Path(Path.cwd())
     match model:
         case "models":
-            return Path(prefix, "models")
+            return Path(Path.cwd(), "models")
```
powderluv, line 80: This is reverting @dan-garvey  patch. Does this work with ckpt_dir provided in the command line ?

Abhishek-Varma: So, if `--ckpt_loc` is being provided to the CLI's exe it works perfectly without touching this part of the code. I tested the CLI for txt2img and it works after this patch too.

This revision concerns WebUI where "`prefix`" will erroneously become only `--ckpt_loc` (irrespective of selecting LoRA/Custom Vae) and it'll concatenate `models/vae` etc to the path.

Let me know if we'd also want to let users access `--ckpt_loc` when invoking WebUI as well - my understanding is that for WebUI we'd let users keep the checkpoints under `models` directory, like earlier, and let them pick from the dropdown list.

powderluv (approved):

PR #1265, opened 2023-03-31
[SD] Add Stable diffusion text2image rest API

Shukla-Gaurav:


apps/stable_diffusion/web/index.py
```diff
@@ -1,209 +1,217 @@
 import os
 import sys
-import transformers
```
powderluv, line 3: we need transformers for the exe to work

PR #1275, opened 2023-04-04
[Web] remove txt2img ui dependencies from png_metadata

m68k-fr:
Revert to Gradio return by list instead of dict, to remove txt2img ui dependencies from png_metadata.

https://gradio.app/blocks-and-event-listeners/#function-return-list-vs-dict


monorimet (approved):
Cool. Was this in response to an issue?

PR #1276, opened 2023-04-04
[SD] Add weight emphasis to the prompts encoder

m68k-fr:
* Implements https://huggingface.co/AlanB/lpw_stable_diffusion_mod on all prompts. (Auto1111 emphasis syntax)
* Few modifications were needed from the original source: To ease the review I tagged them with #SHARK in the source code.
* No support for long prompts for now, as Shark runner is imposing a 77 chunck limit for embeddings.
* Very easy to rollback or switch to the old encoder if it's needed.
* Fully compatible with the old encoder: The new encoder produce exactly the same image as the old one if weights are all set to one.

Tested on txt2img, img2img, inpaint, outpaint and the txt2img command line script.


powderluv (approved):
Wow thank you

PR #1278, opened 2023-04-05
[SD] Fix encoder error for model_max_length not beeing 77

m68k-fr:
Fix the error reported here: https://discord.com/channels/973663919757492264/1092903312740851843
Tested on prompthero/openjourney and confirmed:

pipe.tokenizer.model_max_length is returning 77 instead of 64, ending in a tensor shape error.
I suspect this tokenizer.model_max_length is always returning 77.
To fix the problem the real length was passed as an argument, and we're not relying on this tokeniser property anymore.

Wouldn't it be more appropriate to fix the property itself ?
I check the code and noticed than it is also used in the LoRA trainer.


powderluv:
lets land this as a hotfix and evaluate the right fix.

Issue #1268, opened 2023-04-02
img2img on 768x512 running out memory


ledroyy:
Prompts on 768x512 or 512x768 will make one image but trying again will show this error "VK_ERROR_OUT_OF_DEVICE_MEMORY". If I try changing the model or size it crashes and closes shark. 
img2img on 512x512 has no errors

Version 648 - GPU RX 5700 XT 8G, 16G ram

ledroyy:
Still running out of memory but at least I can change models and size without having to re-open shark everytime. Might have to change to 640x512 or change heap size 

```
Traceback (most recent call last):
  File "gradio\routes.py", line 393, in run_predict
  File "gradio\blocks.py", line 1108, in process_api
  File "gradio\blocks.py", line 929, in call_function
  File "anyio\to_thread.py", line 31, in run_sync
  File "anyio\_backends\_asyncio.py", line 937, in run_sync_in_worker_thread
  File "anyio\_backends\_asyncio.py", line 867, in run
  File "gradio\utils.py", line 490, in async_iteration
  File "apps\stable_diffusion\scripts\img2img.py", line 253, in img2img_inf
  File "apps\stable_diffusion\src\pipelines\pipeline_shark_stable_diffusion_img2img.py", line 178, in generate_images
  File "apps\stable_diffusion\src\pipelines\pipeline_shark_stable_diffusion_utils.py", line 202, in produce_img_latents
  File "apps\stable_diffusion\src\pipelines\pipeline_shark_stable_diffusion_utils.py", line 103, in load_unet
  File "apps\stable_diffusion\src\models\model_wrappers.py", line 586, in unet
  File "apps\stable_diffusion\src\utils\utils.py", line 611, in fetch_vmfb
  File "apps\stable_diffusion\src\utils\utils.py", line 602, in load_vmfb
  File "shark\shark_inference.py", line 207, in load_module
  File "shark\iree_utils\compile_utils.py", line 333, in load_flatbuffer
  File "shark\iree_utils\compile_utils.py", line 309, in get_iree_module
  File "iree\runtime\system_api.py", line 255, in add_vm_module
  File "iree\runtime\system_api.py", line 252, in add_vm_modules
[...]

PR #1283, opened 2023-04-07
Unload clip and workaround the memory leak for ondemand

jinchen62:
- Enable unloading clip on demand
- Workaround the memory leak of unloading models (It still keeps something in memory after unloading models and using gc would cause error. I think something is still referenced so it can't be cleaned. The workaround is creating the SD pipeline everytime if on demand.)
- Disable unloading models for outpainting becuase it would load and unload models for each direction

### DIFF
--- a/apps/stable_diffusion/scripts/img2img.py
+++ b/apps/stable_diffusion/scripts/img2img.py
@@ -170,7 +170,8 @@ def img2img_inf(
         use_stencil=use_stencil,
     )
     if (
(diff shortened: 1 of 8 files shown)


jinchen62:
@monorimet Could you help to test it?

jinchen62:
should fix #1268

monorimet (approved):
My test results were mixed, the content of this PR seems to work as intended though so LGTM.

PR #1291, opened 2023-04-10
Add auto-tuner to SD apps

yzhang93:


apps/stable_diffusion/scripts/tuner.py
```diff
@@ -0,0 +1,114 @@
...
+    else:
+        raise ValueError(
+            f"{args.annotation_model} is not supported for tuning."
+        )
+
+    return mlir_module, model_name
+
+
+def main():
+    args.use_tuned = False
```
dan-garvey, line 50: why?

apps/stable_diffusion/src/models/model_wrappers.py
```diff
@@ -277,7 +279,7 @@ def forward(self, input):
             extra_args=get_opt_flags("vae", precision=self.precision),
             base_model_id=self.base_model_id,
         )
-        return shark_vae
```
dan-garvey, line 280: does this break functionality elsewhere? do we never expect only one output from these functions?

dan-garvey (changes requested):
Instead of a tuning script, or maybe in addition to the tuning script, could we add this functionality as a pipeline feature so that it is immediately available to web and cli users if they pass a flag? ie --tune_otf as an arg to index.py or main.py?

yzhang93:
> Instead of a tuning script, or maybe in addition to the tuning script, could we add this functionality as a pipeline feature so that it is immediately available to web and cli users if they pass a flag? ie --tune_otf as an arg to index.py or main.py?

That's a good point. The plugin should be in a separate patch though. We'll need to decide where to add the tuning in the webui @Shukla-Gaurav , since the tuning should be model-wise other than pipeline-wise.

apps/stable_diffusion/src/models/model_wrappers.py
yzhang93: In addition, I have tested txt2img and img2img with my patch.

apps/stable_diffusion/scripts/tuner.py
dan-garvey: I think as a rule we should probably print something saying we're overriding a flag if the flag was set to true (even erroneously default in this case)

dan-garvey:
In that case maybe it makes more sense as a function inside of model wrappers that can apply (conditionally) to each model? I'm just imagining the desired functionality, webui or otherwise, is pass a flag (or press a button) and then wait however long, come back, and its running the tuned model. I guess one challenge here is detecting that the model has already been tuned for successive cli invocations/web button generate image calls.

yzhang93:
> In that case maybe it makes more sense as a function inside of model wrappers that can apply (conditionally) to each model? I'm just imagining the desired functionality, webui or otherwise, is pass a flag (or press a button) and then wait however long, come back, and its running the tuned model. I guess one challenge here is detecting that the model has already been tuned for successive cli invocations/web button generate image calls.

I agree. For current stage, we'll have to test tuning internally first. I would prefer to use cli before we expose this to the other users in the webui. Then we'll also have to decide what we'd like to release as a tuner whl package. @powderluv

PR #1295, opened 2023-04-12
Fix low vram mode issues

jinchen62:
- add ondemand back to img2img
- workaround memory leak for batch count

### DIFF
--- a/apps/stable_diffusion/scripts/inpaint.py
+++ b/apps/stable_diffusion/scripts/inpaint.py
@@ -64,6 +64,9 @@ def inpaint_inf(
     args.img_path = "not none"
     args.mask_path = "not none"
     args.ondemand = ondemand
+    if ondemand and batch_count > 1:
+        print("Low VRAM mode currently only supports 1 batch count.")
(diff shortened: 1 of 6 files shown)


monorimet (approved):

quasaray:
These changes seem to have broken outpainting, the config was not updated with `ondemand`

jinchen62:
@quasaray Hi, outpainting doesn't have the support of ondemand yet.

quasaray:
@jinchen62 I encountered a runtime error with the latest prerelease, this PR expects the `ondemand` parameter in utils.Config to be set, but the reference to utils.Config in outpainting was not updated to set this when it is created. 
This has nothing to do with actually using ondemand

jinchen62:
@quasaray Do you encounter this on other modes like img2img?

quasaray:
I have only used text2img and outpainting so far, I only encountered the error on outpainting

jinchen62:
@quasaray thanks for pointing it out, I have push up a PR

PR #1298, opened 2023-04-13
Update list of scheduler available for inferences

m68k-fr:
### Except the Shark GPU one, all schedulers are supported for every type of inference:

* Rename the two scheduler lists used by the Ui:  ~~scheduler_list_txt2img & scheduler_list~~ -> scheduler_list & scheduler_list_cpu_only.
* Scheduler Ui dropdowns defaulted to "EulerDiscrete" instead of "PNDM" as it's now fully supported everywhere.

### Note about the Karras scheduler (KDPM2Discrete):

It's working and generate correct images, but two minor problems were noticed:
* img2img: depending on the number of steps and the denoising strength, it can generate a noise image.
* outpaint: Expending more than one region will only output a partial image.

This is surely a side effect of this scheduler internally doubling the number of steps.
I have not removed it, as it's working in most situations.

### DIFF
--- a/apps/stable_diffusion/scripts/img2img.py
+++ b/apps/stable_diffusion/scripts/img2img.py
@@ -34,16 +34,11 @@ def main():
         args.scheduler = "DDIM"
         args.hf_model_id = "runwayml/stable-diffusion-v1-5"
         image, args.width, args.height = resize_stencil(image)
-    elif args.scheduler != "PNDM":
-        if "Shark" in args.scheduler:
-            print(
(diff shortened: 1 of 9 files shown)


monorimet:
@Shukla-Gaurav I know we needed PNDM as default/fallback scheduler for img2img -- this will affect plugins etc. so we should test extensively and maybe enable more in the blender API?

monorimet (approved):
LGTM but needs extensive testing. We can merge and rollback if there are issues with img2img.

PR #1300, opened 2023-04-13
[Web] Style Fixes for Gradio V3.25.0

m68k-fr:
* Fix the Import PNG metadata box height (it's now controlled from the parent div)
* Add a default body background color.
* Gallery: Replace deprecated grid style option.
* Gallery: Add object_fit style property to get correct image ratio no matter the size.


monorimet (approved):
Thanks! looks good to me

PR #1316, opened 2023-04-19
WIP: Allow SD model wrapper to be called for sharktank gen.

monorimet:


powderluv (approved):

apps/stable_diffusion/src/models/model_wrappers.py
```diff
@@ -666,3 +666,44 @@ def controlnet(self):
             return compiled_stencil_adaptor
         except Exception as e:
             sys.exit(e)
+
+	 def __call__(self):
+			# Step 1:
```
jinchen62, line 671: seems the format is not right

monorimet: yeah it's not right for a few reasons. moving to draft

PR #1320, opened 2023-04-19
Don't dump image data on img2img api call.

monorimet:
Currently when using the SD API, we print the entire input data dict which dumps the very long, non-human readable input image data.
Simplify the print statement such that prompt, negative prompt, and seed are printed exclusively.


Shukla-Gaurav (approved):
LGTM! thanks.

Shukla-Gaurav (approved):

PR #1331, opened 2023-04-21
Add StableLM model

vivekkhandelwal1:


apps/language_models/scripts/stablelm.py
```diff
@@ -0,0 +1,207 @@
...
+    use_tracing=False,
+    verbose=False,
+)
+
+bytecode_stream = BytesIO()
+module.operation.write_bytecode(bytecode_stream)
+bytecode = bytecode_stream.getvalue()
+
+shark_module = SharkInference(
+    mlir_module=bytecode, device="cuda", mlir_dialect="tm_tensor"
```
powderluv, line 196: Can we please make this device an argument to the script?

powderluv (approved):
Can we also add a flag to save/load MLIR files and also to skip torch-mlir compile and IREE compile if the vmfb exists ?

PR #1338, opened 2023-04-22
Add device to scheduler model names

nirvedhmeshram:
This changes the scheduler model names from something like  `euler_scale_model_input_1_512_512fp16` to `euler_scale_model_input_1_512_512_vulkan_fp16`
It is useful if you are going to use multiple backends (e.g cuda/vulkan) so that they dont try to use each others .vmfb


monorimet (approved):
LGTM. Thanks for the fix.

PR #1357, opened 2023-04-26
Add support of using civitai model download url

jinchen62:


apps/stable_diffusion/web/ui/txt2img_ui.py
```diff
@@ -290,9 +293,9 @@ def txt2img_api(
                             )
                             hf_model_id = gr.Textbox(
                                 elem_id="hf_model_id",
-                                placeholder="Select 'None' in the Models dropdown on the left and enter model ID here e.g: SG161222/Realistic_Vision_V1.3",
+                                placeholder="Select 'None' in the Models dropdown on the left and enter model ID here e.g: SG161222/Realistic_Vision_V1.3, https://civitai.com/api/download/models/15236",
```
powderluv, line 296: can this just be 15236 ? so they don't have to add https://civitai.com/api/download/models

powderluv: we can call it model number

powderluv (approved):

PR #1380, opened 2023-05-02
Add example JAX MiniLM inference

sogartar:


shark/examples/shark_inference/minilm_jax.py
```diff
@@ -0,0 +1,57 @@
...
+    model_mlir.operation.write_bytecode(file=byte_stream)
+    return byte_stream.getvalue()
+
+
+sample_input = get_sample_input()
+mlir = export_to_mlir(sample_input)
+
+# Compile and load module.
+shark_inference = SharkInference(
+    mlir_module=mlir, mlir_dialect="mhlo", device="cpu"
```
powderluv, line 52: can we parameterize the device so it can run on any supported backend

sogartar: Done.

powderluv (approved):

PR #1382, opened 2023-05-02
Convert original vae to diffusers

jinchen62:


powderluv:
Does this still work with SD2.1 VAE since we link to the 1.4 version ?

powderluv (approved):

PR #1390, opened 2023-05-04
[SD] Fix stable LM chatbot

Shukla-Gaurav:


apps/stable_diffusion/web/index.py
```diff
@@ -136,12 +136,12 @@ def register_button_click(button, selectedid, inputs, outputs):
                 upscaler_web.render()
             with gr.TabItem(label="Model Manager", id=5):
                 model_web.render()
+            with gr.TabItem(label="Chat Bot", id=7):
```
jinchen62, line 139: Though it might not matter, here is 6 in order.

PR #1400, opened 2023-05-05
Base script for vicuna

PhaneeshB:
- Add script to run Vicuna fp32 cuda for a single reply chatbot
- both mlir and vmfb will be compiled on device and will require a large amount (>250G) of mem space


powderluv:
Please get lint to work

PR #1408, opened 2023-05-10
added vicuna demo

Eliasj42:


powderluv (approved):
We should integrate it eventually

PR #1417, opened 2023-05-10
fixed spacing issue with chat-bot

Eliasj42:


apps/language_models/scripts/sharded_vicuna_fp32.py
```diff
@@ -621,39 +620,36 @@ def get_sharded_model():
...
+            tokens.append(new_token)
+            original_input_ids.append(new_token)
+            input_ids = [new_token]
+
+        for i in range(len(tokens)):
+            if type(tokens[i]) != int:
+                tokens[i] = int(tokens[i][0])
+        print(tokenizer.decode(new_sentence))
+        prompt_history += f"\n{tokenizer.decode(new_sentence)}\n"
+        if len(tokenizer(prompt_history).input_ids) > 5000:
```
raikonenfnu, line 653: Is it actually 5000? If it is, why is the padding length only til 137?

Eliasj42: there's no padding anymore, still some legacy code though.  I tried to remove it with this patch.  I think we can just remove this limit entirely

raikonenfnu: Wait is the vmfb dynamic or static? it certainly would have some limits no?

apps/language_models/scripts/sharded_vicuna_fp32.py
```diff
@@ -621,39 +620,36 @@ def get_sharded_model():
         )
         prompt = prompt_history.strip()
         input_ids = tokenizer(prompt).input_ids
+        tokens = input_ids
         prompt = print("Robot:", end=" ")
-        new_sentence = ""
-        for _ in range(200):
+        new_sentence = []
+        for iteration in range(200):
```
raikonenfnu, line 626: NIT: Can we set this as a global-ish variable/constant i.e `MAX_LEN_TOKEN` for better readability. :)

Eliasj42: you mean 200?  would it be better to change it to while True?

raikonenfnu:
LGTM!

raikonenfnu (approved):

PR #1461, opened 2023-05-24
Enable cpu f16 dtype tracing for the vicuna model.

pashu123:


pashu123:
This is generic and applies to Stable diffusion and LLMs. @powderluv @PhaneeshB

PR #1470, opened 2023-05-28
Web/UI: Add an Output Gallery tab for SD

one-lithe-rune:
_I realise that this is probably a bit much for a first contribution to a project, and may not fit with where you want the UI to go, but I needed it myself and was writing it anyway. Therefore PR upstream._  

## Motivation

My Chromebook is in the kitchen, my computer with the graphics card is not. So I run the webui on the computer, and use then browse in from the Chromebook. 

This has some problems, mostly due not having access to the filesystem on which the application is running.

* I can only see images for the last generation batch, rather than all the images I've made that are sitting under the output directory or where I've archived them.
* I can't look at the headers of existing .PNGs on the filesystem to see parameter info.
* I have some JPEG images where I have parameter info saved in the EXIF tags, these have a similar problem.
* Nor can I drag or load those .PNGs into txt2img to read in the parameters for the next generation batch, because that takes images from the filesystem of the machine the browser is on.

This makes the workflow fairly annoying.

## Changes

Adds an new Output Gallery tab to the ui/webui.

### Screenshot

outgallery-demo

### Features

* Subdirectory select dropdown, that lists subdirectories at any depth below the `<output_dir>/generated_imgs` directory,
* Large, full height, gallery area displaying the images in the selected subdirectory. Shows nod logo when no images are in the selected subdirectory.
* Slider that changes the number of columns of images that the gallery displays from between 1 to 16 columns (defaults to 4).
* Expandable parameter info panel showing any generation parameters saved in the file of the selected image for PNGs, alternatively the image's EXIF data for JPEGs
* Send to buttons for txt2img, img2img, inpaint, outpaint and upscaler.
* Auto update of gallery and gallery label (to show generation status), when a new image is generated by any of the stable diffusion tabs, and is outputted to the currently selected subdirectory.
* Command line option `--output_gallery / no-output_gallery` for enabling and disabling the output gallery (defaults to enabled)
* Command line option `--output_gallery_followlinks / --no-output_gallery_followlinks` for following symlinks when getting entries for the subdirectory list (defaults to off, as Python os.walk doesn't check for circular references if following symlinks)

## Known Problems/Concerns

* I wrote this against gradio 3.22.0, since that's the version that is pinned in `requirements.txt` and thus what I get when I do `setup_venv`. But that version complains about existing code's usage of gallery styling parameters that don't exist yet in that release, and since one of the styling parameters is for specifying the number of columns in a gallery, coding against this version using a different parameter may have been a mistake on my part.
* Gradio doesn't support changing the the number of columns in a gallery at runtime (it's in the Gradio 4.0 bucket list AFAIR). You can only set when the gallery is first defined. Therefore I'm using JavaScript shenanigans to make the column number slider work. I considered skipping that feature but doing so made things substantially less useful for me.
* I'm occasionally getting `OSError: [WinError 121] The semaphore timeout period has expired` when I have a bad wifi connection to the computer from the Chromebook, and working with subdirectories that have many hundreds of images in them especially on a slow external drive, so it's very likely I/O related but I haven't been able to track down the exact trigger as it's failing somewhere deep in the bowels of gradio's use of websockets, and I'm lost there.
* Reading parameter info from the .CSV or .JSON files. It doesn't do it...yet.
* I've only tested (manually) the server side on windows, and the browser side extensively on Firefox on Windows, and Chrome on Chromebook.
* I didn't find much about Contributing either in the project or on Discord (I may have missed it!) so I don't know what you might need from me outside of the code.
* This is the squashed version of the branch, the full version is [here](https://github.com/nod-ai/SHARK/compare/main...one-lithe-rune:SHARK:gallery-tab) I don't know which you prefer.

### DIFF
--- a/apps/stable_diffusion/web/ui/upscaler_ui.py
+++ b/apps/stable_diffusion/web/ui/upscaler_ui.py
@@ -1,8 +1,6 @@
-from pathlib import Path
 import os
 import torch
 import time
-import sys
 import gradio as gr
 from PIL import Image
 import base64
@@ -15,18 +13,17 @@ from apps.stable_diffusion.web.ui.utils import (
     get_custom_model_files,
     scheduler_list_cpu_only,
     predefined_upscaler_models,
-    cancel_sd,
 )
+from apps.stable_diffusion.web.utils.common_label_calc import status_label
 from apps.stable_diffusion.src import (
     args,
     UpscalerPipeline,
     get_schedulers,
     set_init_device_flags,
     utils,
-    clear_all,
     save_output_img,
 )
-
+from apps.stable_diffusion.src.utils import get_generated_imgs_path
 
 # set initial values of iree_vulkan_target_triple, use_tuned and import_mlir.
 init_iree_vulkan_target_triple = args.iree_vulkan_target_triple
@@ -208,7 +205,7 @@ def upscaler_inf(
         generated_imgs.append(high_res_img)
         seeds.append(img_seed)
         global_obj.get_sd_obj().log += "\n"
-        yield generated_imgs, global_obj.get_sd_obj().log
+        yield generated_imgs, global_obj.get_sd_obj().log, status_label("Image-to-Image", current_batch+1, batch_count, batch_size)
 
     total_time = time.time() - start_time
     text_output = f"prompt={args.prompts}"
@@ -220,7 +217,7 @@ def upscaler_inf(
     text_output += global_obj.get_sd_obj().log
     text_output += f"\nTotal image generation time: {total_time:.4f}sec"
 
-    yield generated_imgs, text_output
+    yield generated_imgs, text_output, ""
 
 
 def decode_base64_to_image(encoding):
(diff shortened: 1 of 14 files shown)


powderluv:
Thank you for your contribution. Please run the black linter.  We should unpin from 3.22 at somepoint (I forget why we pinned @PhaneeshB  / @Shukla-Gaurav ) may know. 

But happy to land the PR and iterate once linter is fixed (and there are no regressions).

one-lithe-rune:
Ah, spotted it needed linting with flake8, but didn't notice black. Done.

apps/stable_diffusion/web/ui/upscaler_ui.py
```diff
@@ -15,18 +13,17 @@
     get_custom_model_files,
     scheduler_list_cpu_only,
     predefined_upscaler_models,
-    cancel_sd,
```
powderluv, line 18: Why remove? Don't we want an option to cancel ?

one-lithe-rune: I'll put them back, I removed them because it didn't look like anything was using them.

one-lithe-rune: Hmm...We might want an option to cancel but at the moment `cancel_sd` isn't hooked up to the stop button on upscaler tab, this is the current code:

```python
        stop_batch.click(
            fn=None, cancels=[prompt_submit, neg_prompt_submit, generate_click]
        )
``` 

Which is probably why my editor flagged it as not used. Do you want me to change that on this PR or should I do it on another one?

one-lithe-rune: Import is back in as of my latest commit. Left the `stop_batch` click handler for now.

powderluv:
@monorimet can you take a quick look too. We can land once CI passes.

monorimet (approved):
Looks good to me. We should merge it and get the community's opinion, but to me, these are direct QOL improvements. Thanks for the contribution.

monorimet (approved):

powderluv (approved):

PR #1475, opened 2023-05-31
Add to and tweak vulkan configuration environments.

Foxlum:
Adds **[VK_KHR_shader_integer_dot_product](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VK_KHR_shader_integer_dot_product.html)** (a vulkan extension exposing the SPIR-V extension **[SPV_KHR_integer_dot_product](https://htmlpreview.github.io/?https://github.com/KhronosGroup/SPIRV-Registry/blob/master/extensions/KHR/SPV_KHR_integer_dot_product.html)**) for some compatible Vulkan environments which is included in core Vulkan 1.3, it can have the potential to improve (most commonly) int8 inferencing/training speeds and better performance as dot products may be more optimal and most importantly accelerated on capable hardware.

This does likely require some changes to the backend to see any full use out of if possible, and I may have written this commit in an erroneous way that may not properly define capabilities or create problematic issues across other architectures. (This commit defines it for Nvidia, AMD, and Intel GPUs, with no apparent compatibility issues at least for a reasonable time range of GPUs based off of [GPUinfo.org](https://vulkan.gpuinfo.org/) data.)

(Edit: forgot to mention, explicitly turned off ShaderFloat16 on Nvidia Pascal, as it doesn't seem to support them.)

### DIFF
--- a/shark/iree_utils/vulkan_target_env_utils.py
+++ b/shark/iree_utils/vulkan_target_env_utils.py
@@ -243,7 +243,7 @@ def get_vulkan_target_capabilities(triple):
             cap["coopmatCases"] = [
                 "mSize = 16, nSize = 16, kSize = 16, aType = f16, bType = f16, cType = f16, resultType = f16, scope = #vk.scope<Subgroup>"
             ]
-        
+
         if product == "rx5700xt":
             cap["storagePushConstant16"] = False
             cap["storagePushConstant8"] = False
--- a/shark/iree_utils/vulkan_target_env_utils.py
+++ b/shark/iree_utils/vulkan_target_env_utils.py
@@ -117,7 +117,8 @@ def get_extensions(triple):
 
     if get_vendor(triple) == "NVIDIA" or arch == "rdna3":
         ext.append("VK_NV_cooperative_matrix")
-
+    if get_vendor(triple) == ["NVIDIA", "AMD", "Intel"]:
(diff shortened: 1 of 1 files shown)


powderluv:
Thanks for the PR. Please run the black linter. Also we are moving to the vulkan env but the old way of doing it is still there (Here is an example https://github.com/openxla/iree/pull/12911/files ). Looks like it has been added but we have to verify it is enabled for the cards we care about here: https://github.com/openxla/iree/blob/main/compiler/src/iree/compiler/Dialect/Vulkan/Utils/TargetTriple.cpp

Foxlum:
> Thanks for the PR. Please run the black linter. Also we are moving to the vulkan env but the old way of doing it is still there (Here is an example https://github.com/openxla/iree/pull/12911/files ). Looks like it has been added but we have to verify it is enabled for the cards we care about here: https://github.com/openxla/iree/blob/main/compiler/src/iree/compiler/Dialect/Vulkan/Utils/TargetTriple.cpp 
> 

So I should black lint what I have, and adjust the configuration within the .cpp and other appropriate locations?

powderluv:
yes please. the lint commands are in https://github.com/nod-ai/SHARK/actions/runs/5128191414/jobs/9225128359?pr=1475 

You may not have to adjust the cpp (just verify it matches the changes you are making).

powderluv (approved):

PR #1486, opened 2023-06-05
Replace RDNA2 SD2.1 Unet lowering configs.

monorimet:
New 512x512 and 768x768 unet tunings (768 speeds up a little with this config but not by much. Tuning was disabled for 768^2 on rdna2 previously.)


powderluv (approved):

powderluv (approved):

powderluv:
We should also add 512x768 and 768x512

monorimet:
We got a few more it/s on 512x512 though. Up to ~9.5 on my 6900xt.

PR #1491, opened 2023-06-05
Add .csv and .json parameter info to output gallery

one-lithe-rune:
### Changes

* Gives the output gallery the ability to read parameter information from the .csv and .json files. Priority for the source for parameter info when selecting an image is now:  PNG -> Json -> Csv -> Exif.
* Refactors the resulting and existing UI related metadata wrangling functions into an apps.stable_diffusion.web.utils.metadata module.

### Concerns

* All the .csv's I have, don't have headers, and they have differing formats as more parameters have been added over time, so I'm guessing at what the format is based on the number of fields, and then matching things up based on definitions I know about/had kicking around.
* Code to deal with the fact that JSON doesn't include filenames but CSV does is kind of ugly.
* Text2Image can still only import parameters from .PNGs for now, and there's nothing to indicate/prevent you from using sendto when it's not a .PNG. It'll just do nothing that case but it's confusing.


one-lithe-rune:
Pushed a rebase, so should now be up to date with #1493.

PR #1493, opened 2023-06-05
Fix output gallery not using shark_tmp and thus eating disk space

one-lithe-rune:
### Motivation

The code that makes sure that gradio images end up in `shark_tmp` rather than the system's or user's generic `tmp` directory doesn't work for the output gallery. So those temporary files it generates never get removed on application restart.

Since the output gallery shows lots of images and there are multiple temporary files per pic inside a directory per image, this gets nasty rather quickly. (25GiB of images and 1.6 million files for me). I think this must have something to do with the output gallery working with filenames rather than PIL images which is what the current code is setup to deal with.

### Changes

- Sets the output gallery tab's gallery component to it its temp files in `shark_tmp/galleries` 
- does `shutil.rmtree` on that `galleries` subdirectory on startup.

### Problems/Concerns

- I don't think the setup for the other galleries is working entirely correctly, even though the images all go into `shark_tmp` you end up with duplicates of the same image which I don't get with the output gallery. I haven't attempted to fix this, it might just a peculiarity of how gradio handles PIL images.
- This will probably conflict with my other PR #1491, not entire sure how we want to handle that. Of the two, this is more important.


monorimet (approved):
LGTM. Thanks for contributing the fix.

monorimet:
Merging despite unrelated CI crash on CPU job. Stable diffusion tests succeeded.

PR #1503, opened 2023-06-06
Prevent having no cuda devices for LLM breaking the UI

one-lithe-rune:
### Motivation/Changes

https://github.com/nod-ai/SHARK/commit/8c219604861a2b9a349fe645d91e253022256f93 broke UI/Web on startup as it doesn't take into account the fact that it is possible for ones machine to not have any cuda devices.

This handles that case.

### Possible Problems

* Since I don't have any cuda devices, I haven't tested whether this does what we want when you do, although it should if https://github.com/nod-ai/SHARK/commit/8c219604861a2b9a349fe645d91e253022256f93 was correct there.
* Blocks both vircuna and stabilitylm if no cuda. I don't know whether this is correct for stabilitylm.


powderluv (approved):
@dan-garvey

PR #1504, opened 2023-06-07
Add UNet512

gpetters94:
The frontend work still needs some cleanup I think, and I need to check if it's truncating the prompts still, but otherwise I think this is good to go.

### DIFF
--- a/apps/stable_diffusion/src/models/model_wrappers.py
+++ b/apps/stable_diffusion/src/models/model_wrappers.py
@@ -163,7 +163,7 @@ class SharkifyStableDiffusionModel:
 
     def get_extended_name_for_all_model(self):
         model_name = {}
-        sub_model_list = ["clip", "unet", "stencil_unet", "vae", "vae_encode", "stencil_adaptor"]
+        sub_model_list = ["clip", "unet", "unet512", "stencil_unet", "vae", "vae_encode", "stencil_adaptor"]
         index = 0
         for model in sub_model_list:
             sub_model = model
@@ -415,7 +415,7 @@ class SharkifyStableDiffusionModel:
         )
         return shark_cnet, cnet_mlir
 
-    def get_unet(self):
+    def get_unet(self, use_large=False):
         class UnetModel(torch.nn.Module):
             def __init__(self, model_id=self.model_id, low_cpu_mem_usage=False, use_lora=self.use_lora):
                 super().__init__()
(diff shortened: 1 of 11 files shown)


apps/stable_diffusion/src/pipelines/pipeline_shark_stable_diffusion_utils.py
```diff
@@ -203,7 +221,10 @@ def produce_img_latents(
         latent_history = [latents]
         text_embeddings = torch.from_numpy(text_embeddings).to(dtype)
         text_embeddings_numpy = text_embeddings.detach().numpy()
-        self.load_unet()
+        if(text_embeddings.shape[1] <= 64):
```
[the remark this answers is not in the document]
gpetters94: We don't need to, since all unloading does is

```
del x
x = None
```

which works fine even if it's not loaded. I can change it for clarity's sake though.

apps/stable_diffusion/src/models/model_wrappers.py
```diff
@@ -502,6 +509,13 @@ def forward(self, latent, timestep, text_embedding, noise_level):
         unet = UnetModel(low_cpu_mem_usage=self.low_cpu_mem_usage)
         is_f16 = True if self.precision == "fp16" else False
         inputs = tuple(self.inputs["unet"])
+        if(sequence_length > 64):
```
one-lithe-rune, line 512: Should this be `if(sequence_length > arg.max_length` or something equivalent if args.maxlength in passed in? 

Same for other places where `64` is hardcoded.

That's not a rhetorical question, I don't actually know :)

gpetters94: Oh, good point. I should probably just pass a bool there, something like `use_large` or `use_512`.

cstueckrath:
this currently breaks ui
```
  File ".\SHARK\apps\stable_diffusion\web\ui\txt2img_ui.py", line 183, in txt2img_inf
    out_imgs = global_obj.get_sd_obj().generate_images(
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: Text2ImagePipeline.generate_images() missing 1 required positional argument: 'max_embeddings_multiples'
```

monorimet:
A few considerations that should be accounted for in this implementation:

- [ ] Ideally the switch to seqlen512 unet should happen automatically if the prompt embeds are longer than 64. This "switch" can happen after CLIP encodes the prompt into the embeddings input for UNet -- either `max_embeddings_multiples` was utilized and the output batches of CLIP model are concatenated (then we need to use the 512 unet), or the prompt was short enough that (even though `max_embeddings_multiples` > 1) the output of CLIP is not longer than 64 so no need to change the output. This is how the existing implementation for the CLIP workaround functions. I think a better way might be to keep the default `max_embeddings_multiples` at 5 -- if the prompt is long enough to use the multiples, we choose large seqlen unet automatically, and if not, use the seqlen64 unet. Then there's no need for the user to supply `--max_embeddings_multiples` manually.

- [ ] If users compile a .vmfb for seqlen 64 unet (via webui for example), the SD app will try to use that UNet again even if it should compile a larger one. The user will then get the input dimension mismatch error. You can reproduce this issue by running the webui with `python index.py --max_embeddings_multiples=5` and with the change to `txt2img_ui.py` to pipe the `args.max_embeddings_multiples` to `generate_images()`

monorimet (changes requested):
This is good progress and I'm glad to see it making a difference on my local tests. There are some issues/considerations that should be made before merging.

apps/stable_diffusion/web/ui/txt2img_ui.py
```diff
@@ -193,6 +193,7 @@ def txt2img_inf(
             dtype,
             args.use_base_vae,
             cpu_scheduling,
+            args.max_embeddings_multiples,
```
monorimet, line 196: This should happen earlier when the Text2ImagePipeline is initialized. 
As is the seqlen64 unet/CLIP is selected and the prompt is truncated before `max_embeddings_multiples` is set to 5 when you run via the webui.

monorimet (changes requested):

apps/stable_diffusion/web/ui/txt2img_ui.py
gpetters94: Are you sure? I've verified it's using unet512 from the webui and it's producing identical images.

monorimet: Yeah thanks for taking a look. May have just needed clearing cache etc.

Good to go now.

monorimet (approved):
LGTM

PR #1509, opened 2023-06-08
fix Cuda OOM vic cli unsharded

PhaneeshB:


apps/language_models/src/pipelines/vicuna_pipeline.py
```diff
@@ -402,7 +408,8 @@ def compile(self):
         else:
             # get first vic
             # TODO: Remove after testing to avoid memory overload
-            fvic_shark_model = self.compile_first_vicuna()
```
dan-garvey, line 405: how is this functionality replaced?

dan-garvey (changes requested):

dan-garvey (approved):

PR #1511, opened 2023-06-08
Add a model config generator

nithinsubbiah:
Model config generator takes a PyTorch model as input and generates a JSON file with model layers and other propperties that define sharding on a particular hardware.

### DIFF
--- a/apps/language_models/src/model_wrappers/vicuna_model.py
+++ b/apps/language_models/src/model_wrappers/vicuna_model.py
@@ -237,3 +237,25 @@ class SecondVicuna(torch.nn.Module):
             return_vals.append(item[0])
             return_vals.append(item[1])
         return tuple(return_vals)
+
+
+class CombinedModel(torch.nn.Module):
+    def __init__(
+        self,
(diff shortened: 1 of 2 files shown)


apps/language_models/src/model_wrappers/vicuna_model.py
```diff
@@ -237,3 +235,21 @@ def forward(
             return_vals.append(item[0])
             return_vals.append(item[1])
         return tuple(return_vals)
+
+class CombinedModel(torch.nn.Module):
+
+    def __init__(self):
+        super().__init__()
+        self.first_vicuna = FirstVicuna("TheBloke/vicuna-7B-1.1-HF")
+        self.second_vicuna = SecondVicuna("TheBloke/vicuna-7B-1.1-HF")
```
powderluv, line 244: this can't be hardcoded right ?

nithinsubbiah: fixed

powderluv:
lint please

monorimet (approved):
LGTM.

Issue #1522, opened 2023-06-12
(SD) (RDNA3) Unet failes to compile for SD1.4 on windows CI


monorimet:
```
C:\actions-runner\w\SHARK\SHARK\shark.venv\Scripts\python.exe apps/stable_diffusion/scripts/txt2img.py --device=vulkan --prompt="cyberpunk forest by Salvador Dali" --negative_prompts="" --seed=42 --import_mlir --output_dir=C:\actions-runner\w\SHARK\SHARK\test_images\CompVis/stable-diffusion-v1-4 --hf_model_id=CompVis/stable-diffusion-v1-4 --no-use_tuned --no-progress_bar
failed to generate image for this configuration
C:\actions-runner\w\SHARK\SHARK\shark.venv\Lib\site-packages\torch\jit\_check.py:172: UserWarning: The TorchScript type system doesn't support instance-level annotations on empty non-base types in `__init__`. Instead, either 1) use a type annotation in the class body, or 2) wrap the type in `torch.jit.Attribute`.

  warnings.warn("The TorchScript type system doesn't support "

WARNING: [Loader Message] Code 0 : windows_read_data_files_in_registry: Registry lookup failed to get layer manifest files.

WARNING: [Loader Message] Code 0 : windows_read_data_files_in_registry: Registry lookup failed to get layer manifest files.

WARNING: [Loader Message] Code 0 : windows_read_data_files_in_registry: Registry lookup failed to get layer manifest files.

shark_tank local cache is located at C:\Users\anush\.local/shark_tank/ . You may change this by setting the --local_tank_cache= flag

Found device AMD Radeon RX 7900 XTX. Using target triple rdna3-7900-windows.

Tuned models are currently not supported for this setting.

No vmfb found. Compiling and saving to C:\actions-runner\w\SHARK\SHARK\euler_scale_model_input_1_512_512_vulkan_fp16.vmfb

Configuring for device:vulkan://00000000-0300-0000-0000-000000000000

Using target triple -iree-vulkan-target-triple=rdna3-7900-windows from command line args

Saved vmfb in C:\actions-runner\w\SHARK\SHARK\euler_scale_model_input_1_512_512_vulkan_fp16.vmfb.

No vmfb found. Compiling and saving to C:\actions-runner\w\SHARK\SHARK\euler_step_1_512_512_vulkan_fp16.vmfb

Configuring for device:vulkan://00000000-0300-0000-0000-000000000000

[...]

PR #1523, opened 2023-06-12
Fix (?) indents in add_upcast in shark importer

one-lithe-rune:
This addresses (https://github.com/nod-ai/SHARK/issues/1522)

* The two with blocks in add_upcast appear to be underindented making SD 1.4 break on rdna3 per @monorimet, I've pushed them out one more tab, and then everything appears to work again.

Disclaimer: I have no idea what this code does other than the indents looking wrong and that's where it breaks.


monorimet (approved):
@pashu123 FYI (sorry for the late ping) let me know if this was intentional and we can revert.
This looks right to me. Thanks for the help.

Issue #1529, opened 2023-06-13
Web/UI Start is extremely slow if you've used the output gallery a lot in the previous session


one-lithe-rune:
The process of cleaning up the gallery files create from `shark_tmp` can be very slow, and its not obvious what is happening while it is doing this.

This is a nasty combination of Gradio creating lots of zero length and duplicate files in many subdirectories for galleries, especially in the 3.22.0 pinned version we are using (https://github.com/gradio-app/gradio/pull/4256 supposedly fixes this in 3.32.0 and later) and python's `shutil.rmtree` being very slow for various reasons.

I'll add something that prints out what is happening, and try a different method of removing the subdirectories and files, but I need to check that actually helps before going ahead with a PR.

Putting this issue up in case it doesn't, or doesn't help enough, and to record the gradio issue.

one-lithe-rune:
Okay, looks like the `shutil.rmtree` slowness is not a thing in recent python versions.

The problem is mostly down to gradio creation of those zero length files, which the 3.22.0 version creates for every image in a gallery every time it is loaded or updated.

powderluv:
yikes. thanks for digging into it. We pinned to 3.22 for some transient breakage. We can try to unpin to the latest.

one-lithe-rune:
PR with a workaround, on the assumption we need to keep 3.22 pinned: https://github.com/nod-ai/SHARK/pull/1531

one-lithe-rune:
I was using a more recent version of gradio recently to try and get gradio's version HMR working (I kinda sorta got it working but I decided it wasn't very useful in a multi-tab environment), and did cause some breakage, most immediately the randomize buttons stopped working, so if we did upgrade it would need some attention.

powderluv:
3.22 has some security vulnerabilities so we have to update now . @Shukla-Gaurav fyi in case you can do it as part of your changes

one-lithe-rune:
Closing now https://github.com/nod-ai/SHARK/pull/1548 has been merged.

PR #1531, opened 2023-06-13
UI/Web Improve output gallery temp file handling

one-lithe-rune:
Partially addresses https://github.com/nod-ai/SHARK/issues/1529

### Motivation

I wondered why my startup was taking more than 10 minutes this evening.

It turned out that gradio < 3.32.0 writes a pointless extra zero length temporary file for every image in the output gallery, every time that gallery is updated. Since the output gallery is updated for every image generated, and for every time you switch its source directory, that's a lot of nonsense zero length files multiplying out of control.

And it took just that long to delete them all during the temp file cleanup that happens on Web/UI start.

### Changes

* On startup we now report that cleaning up of temp files is taking place, in case it takes a long time.
* We have the output gallery tab delete any zero length temporary files generated by for its gallery control whenever it needs to update that control with images. This prevents such files multiplying out of control.

### Concerns/Problems

* This is just a workaround. The 'correct' fix would be to upgrade gradio, but assume we have it pinned to 3.32.0 for a reason.


monorimet (approved):
Thanks for this. I'll check out the gradio pin soon.

PR #1535, opened 2023-06-14
Change mlir dialects for tf tests to stablehlo.

monorimet:
Also temporarily removes TF pytests as IREE no longer registers the .mhlo dialect and therefore cannot run .mlir in the mhlo dialect. These will be added back in as soon as sharktank uploads are back on track with stablehlo.


shark/shark_runner.py
```diff
@@ -27,7 +27,7 @@
 # supported dialects by the shark-runtime.
 supported_dialects = {
     "linalg",
-    "mhlo",
+    "mhlo-legacy",
```
powderluv, line 30: should we add `auto`

monorimet: Yeah and it seems like `mhlo-legacy` doesn't actually exist as an iree input type anymore

PR #1539, opened 2023-06-15
Add attention op insertion code

gpetters94:
This replaces the `bmm -> _softmax -> bmm` sequence with `scaled_dot_product_attention`. It accounts for some view-like unary ops in between (view, _unsafe_view, expand) but the list isn't exhaustive - it might need to be expanded for future model support.


pashu123:
Also, try running the standalone fx script and see whether this generates a correct image.

gpetters94:
> Also, try running the standalone fx script and see whether this generates a correct image.

Good idea, I'll post the results once I've gotten them.

gpetters94:
I've verified that the fx graph before and after my code produces the same output for a call to `forward()`. Haven't tested image generation, but it should also be fine. I'll let you know once the IREE changes land how the results of that test go.

pashu123:
> I've verified that the fx graph before and after my code produces the same output for a call to `forward()`. Haven't tested image generation, but it should also be fine. I'll let you know once the IREE changes land how the results of that test go.

Sounds good, according to my speculation this will solve the 768x768 black image generation issue - Please if you could verify the fx graph that would be great.

PR #1541, opened 2023-06-15
Add pretraining pytest script with SHARK dynamo.

monorimet:
Very preliminary work to set up pytests for pretraining with SHARK dynamo.
TODO:

- [ ] Right now most of the module testing block is characterized around the bert-large sample. Generalize pretraining / modularize tests based on .json configs (now more important with testing pretraining -- we can't just use model_name, dynamic, device to represent all models). We either need separate classes in model utils (yucky) or have .json configs that keep all the sizes and arguments etc. etc. for each model in pretrain_models.csv.
- [ ] the torch-mlir dynamo backend wrapper doesn't support passing in extra args, so we need to find a better way to pass device args. etc into the shark backend.
- [ ]  test on linux or find workaround on Windows: `test_models_dynamo.py::SharkModuleTest::test_module_bert_large_uncased_torch_static_cpu_task - RuntimeError: Windows not yet supported for torch.compile`
- [ ] with pytest+unittest it is hard to give custom test parameters as pytest wants to collect tests before parsing command line arguments. simplify test parametrization/ cut out unittest?

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -90,3 +90,8 @@ def pytest_addoption(parser):
         type=int,
         help="Batch size for the tested model.",
     )
+    parser.addoption(
+        "--custom_device",
+        default=None,
+        help="Custom device string to run tests with.",
+    )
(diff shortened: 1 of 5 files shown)


monorimet:
Closing as not planned (would be implemented in SHARK-Turbine)

PR #1542, opened 2023-06-15
changes for tuning

rprasad2:
Changed files: utils.py and sd_annotation.py to allow for use of the tunings generated. Previously passing in arguments with dimesions of 512x768 or 768x512 would set variable use_tuner to false, which wouldn't use the tunings created. I changed that as well as changed the syntax for searching in the cloud to fit the naming convention used on the files.

### DIFF
--- a/apps/stable_diffusion/src/utils/utils.py
+++ b/apps/stable_diffusion/src/utils/utils.py
@@ -297,12 +297,12 @@ def set_init_device_flags():
     if (
         args.precision != "fp16"
         or args.height not in [512, 768]
-        or (args.height == 512 and args.width != 512)
-        or (args.height == 768 and args.width != 768)
+        or (args.height == 512 and args.width not in [512,768])
+        or (args.height == 768 and args.width not in [512, 768])
         or args.batch_size != 1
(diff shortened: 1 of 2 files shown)


apps/stable_diffusion/src/utils/utils.py
```diff
@@ -298,7 +298,8 @@ def set_init_device_flags():
         or args.batch_size != 1
         or ("vulkan" not in args.device and "cuda" not in args.device)
     ):
-        args.use_tuned = False
+       # args.use_tuned = False
```
monorimet, line 301: Instead of commenting this out (which would enable tuning for fp32 and all different sizes and possibly cause issues when we try to find tunings for sizes we don't have), you can change line 296-297 to include your cases more explicitly.
e.g.:
```
or (args.height == 512 and args.width not in [512, 768])
or (args.height == 768 and args.width not in [512, 768])
```

monorimet (changes requested):

powderluv:
Is this with other changes not supposed to be in here ? Also please run the linter

monorimet (approved):
LGTM

PR #1548, opened 2023-06-17
Remove workarounds for fixed gradio tempfile bugs, and update how gradio is pointed at shark_tmp

one-lithe-rune:
### Motivation

The recent update of the Gradio version brought in a lot fixes for its temporary file handling with images and galleries. Since the temporary files have been one of the banes of my life with the output gallery work, now we are more up to date I thought I'd better cleanup the existing workarounds that aren't (shouldn't!) be needed anymore and implement the gradio's current method for telling it where to put temporary files.

### Changes

* Remove workarounds for bugs with temporary files generated for images and galleries in previously used gradio versions.
* Set the GRADIO_TEMP_DIR environment variable, to point to a directory under shark_tmp if it is not already defined. This is now where gradio will put its temporary files.
* Clear previous run temp files by doing shutil.rmtree on the path in GRADIO_TEMP_DIR, rather than the previous methods.

### Testing

* I have checked that everything is going into the designated folder under `shark_tmp` and that nothing is being created in my os temporary file location, during both generation and use of the output gallery. So this does all appear to be working.


powderluv (approved):

powderluv:
Thank you

PR #1557, opened 2023-06-20
fixed a bug where designating device for vicuna didn't work

Eliasj42:
The Vicuna class wasn't passing device variable, now sharded vicuna should use the designated device


PhaneeshB (approved):
LGTM
Thanks!

PR #1559, opened 2023-06-20
Add patch for making compile API work for both MEGABYTE and MiniGPT4

Abhishek-Varma:
Add patch for making compile API work for both MEGABYTE and MiniGPT4

This adds add an optional `device` argument to the compile API.

### DIFF
--- a/shark/examples/shark_inference/mega_test.py
+++ b/shark/examples/shark_inference/mega_test.py
@@ -52,14 +52,22 @@ shark_module, _ = compile_through_fx(
     base_model_id=None,
     model_name="mega_shark",
     precision=None,
-    return_mlir=False,
+    return_mlir=True,
     device="cuda",
 )
 # logits = model(x)
 
+
+def print_output_info(output, msg):
+    print("\n", msg)
+    print("\n\t", output.shape)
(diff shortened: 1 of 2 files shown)


shark/examples/shark_inference/mega_test.py
```diff
@@ -52,14 +52,22 @@ def forward(self, input):
     base_model_id=None,
     model_name="mega_shark",
     precision=None,
-    return_mlir=False,
+    return_mlir=True,
```
Abhishek-Varma, line 55: Because I'm using the compile API which is customised for StableDIffusion - it leads to a key error.
This change would avert the same.

powderluv (approved):

shark/examples/shark_inference/mega_test.py
powderluv: but we should fix the api to not be customized

Abhishek-Varma: Okay, sure - this was just a quick workaround.
I'll work on it and create another API instead which we can use with any other model seamlessly. Will undo this change thereafter.

PR #1567, opened 2023-06-21
Fix bert training

vivekkhandelwal1:
We don't need to use the upstream PyTorch decomposition for the op `aten.embedding_dense_backward`, since now we have the lowering for the same in Torch-MLIR (https://github.com/llvm/torch-mlir/pull/2255).

### DIFF
--- a/shark/backward_makefx.py
+++ b/shark/backward_makefx.py
@@ -15,7 +15,7 @@
 import torch
 from torch._decomp import get_decompositions
 from torch.fx.experimental.proxy_tensor import make_fx
-from torch.nn.utils import _stateless
+from torch.nn.utils import stateless
 
 from torch import fx
 import tempfile
--- a/shark/examples/shark_training/bert_training.py
+++ b/shark/examples/shark_training/bert_training.py
@@ -1,5 +1,5 @@
 import torch
-from torch.nn.utils import _stateless
+from torch.nn.utils import stateless
 from transformers import AutoTokenizer, AutoModelForSequenceClassification
 from shark.shark_trainer import SharkTrainer
 
@@ -33,7 +33,7 @@ inp = (torch.randint(2, (1, 128)),)
 
 def forward(params, buffers, args):
     params_and_buffers = {**params, **buffers}
-    _stateless.functional_call(
+    stateless.functional_call(
         mod, params_and_buffers, args, {}
     ).sum().backward()
     optim = torch.optim.SGD(get_sorted_params(params), lr=0.01)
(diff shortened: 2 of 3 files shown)


pashu123:
@vivekkhandelwal1 I don't feel this is the right approach. Why aren't we using the upstream decomposition?

pashu123:
> > @vivekkhandelwal1 I don't feel this is the right approach. Why aren't we using the upstream decomposition?
> 
> Hi @pashu123, I don't see any specific reason for using only the `upstream PyTorch decomposition`. There are n number of instances where we have a decomposition in the PyTorch (https://github.com/pytorch/pytorch/blob/main/torch/_decomp/decompositions.py) but we are still not using that instead we have added a lowering for the op in Torch-MLIR. Few such examples are aten.tanh_backward, aten.upsample_nearest2d.vec, aten.silu, aten.nll_loss_backward op.
> 
> IIRC @powderluv once mentioned that ideally we should not rely on the upstream decompositions instead should add the lowering in Torch-MLIR. One rationale is that using the upstream decomposition is very much change-sensitive. A small change in the upstream decomposition can lead to multiple failures in the Torch-MLIR compilation. On the other hand, if we have that in Torch-MLIR itself we can avoid these issues and have more stability during the compilation. Also, using or adding a decomposition in the upstream PyTorch was just a workaround to get the op supported faster, otherwise, every op simple/complex can have a decomposition in PyTorch(many of which are already there).
> 
> And if we talk about this specific op, `aten.embedding_dense_backward`, then after this change in the PyTorch [pytorch/pytorch@66e3989](https://github.com/pytorch/pytorch/commit/66e398951a33af12f20dc671cda7128c9258a1ab), it started throwing out this error: image That's why I added the lowering directly in Torch-MLIR instead of relying upon the upstream decompositions, which fixed the issue.
> 
> I would be happy to hear your thoughts and correct myself if I have been wrong somewhere.

The number one reason to rely on upstream Pytorch decomposition is because of the testing infra that it provides - I was having a hard time debugging the `aten.var_mean` validation issue when using the torch-mlir decomposition.
 
Secondly, writing the decomposition in Pytorch is not torch-mlir's problem and is Pytorch problem too when it comes to testing (which is a much bigger community). Also, writing and fixing decomposition in Pytorch is way simpler.

Also, writing the same decomposition again rather than reusing gives a sense of repeatability which should be taken care of from a software development perspective. AFAIK, in the early days @powderluv must have mentioned having in-house when the make_fx thing was experimental but now since it's mostly stable we should be leveraging that.   

My concern is only regarding repeatability and fixing the root cause.

powderluv:
Lets try to land the PR in upstream PyTorch and meanwhile we can unblock ourselves locally.

vivekkhandelwal1:
This PR is not required anymore. Fixed by https://github.com/pytorch/pytorch/commit/12ca224662a630e55a8f6f1f8c5e187d2456051d.

PR #1571, opened 2023-06-22
fp16 fixes for webui

dan-garvey:


powderluv:
Let's land and iterate

PR #1582, opened 2023-06-23
Custom save directory for output images

cqtqlyst:
- Added the option for users to save their output images in a custom directory
- Added .DS_Store to the .gitignore


cqtqlyst:
working on getting this running on ubuntu to see why there are errors, this most recent commit is just syncing with the main repository

PR #1592, opened 2023-06-24
Pin torch-mlir and use local-task in OPT.

monorimet:


powderluv:
you also need to pin in ./setup_venv.ps1 I think

PR #1598, opened 2023-06-26
[WIP] Add int4/int8 vicuna

jinchen62:


shark/shark_importer.py
```diff
@@ -504,33 +505,149 @@ def import_with_fx(
...
+                    torch.ops.aten.slice_backward,
+                    torch.ops.aten.select_backward,
+                    torch.ops.aten.norm.ScalarOpt_dim,
+                    torch.ops.aten.native_group_norm,
+                    torch.ops.aten.upsample_bilinear2d.vec,
+                    torch.ops.aten.split.Tensor,
+                    torch.ops.aten.split_with_sizes,
+                    torch.ops.aten.native_layer_norm,
+                    torch.ops.aten.masked_fill.Tensor,
+                    torch.ops.aten.masked_fill.Scalar,
```
dan-garvey, line 647: Combine any duplicated decomps into a list outside the if statement

dan-garvey (changes requested):

apps/stable_diffusion/src/utils/utils.py
```diff
@@ -106,7 +106,7 @@ def compile_through_fx(
     model,
     inputs,
     extended_model_name,
-    is_f16=False,
+    precision="fp32",
```
Abhishek-Varma, line 109: This wasn't needed.

The function already has `precision` argument!

We shouldn't change StableDiffusion's API anyway!

Abhishek-Varma (commented):
Please update your patch as per the comment - `precision` in StableDiffusion's API won't be needed because it's already there.

apps/stable_diffusion/src/utils/utils.py
jinchen62: Good catch, removed. It's for making `import_with_fx` more general.

apps/language_models/src/pipelines/falcon_pipeline.py
```diff
@@ -167,7 +167,7 @@ def compile_falcon(self):
                 ts_graph = import_with_fx(
                     model,
                     falconCompileInput,
-                    is_f16=self.precision == "fp16",
+                    precision=self.precision,
```
vivekkhandelwal1, line 170: I think making changes in all the existing working models/pipelines won't be a good move especially if you haven't tested this patch for all the related models/pipelines which are changed. We shouldn't merge this patch for now till these pipelines are tested or we are sure about whether all of these changes are required.

vivekkhandelwal1 (changes requested):

apps/language_models/src/pipelines/falcon_pipeline.py
jinchen62: I could just easily adding a precision arg for import_with_fx without changeing is_f16. Dan said we should probably get rid of the is_f16 arg if adding `precision`. I can revert it if you think it's better for merging this patch. I think we should merge this patch soon to unblock the int8 work.

vivekkhandelwal1: The precision arg is already there you can make use of that. And that would also eliminate the need for modifying the other models.

jinchen62: there was no precision arg in import_with_fx

Abhishek-Varma:
`ModuleNotFoundError: No module named 'brevitas_examples'` - @jinchen62 

Can you please resolve such failures we see in CI?

powderluv:
We can add Brevitas to the requirements.txt with a git install

apps/language_models/src/pipelines/falcon_pipeline.py
jinchen62: reverted and added a todo for getting rid of is_f16

.gitignore
```diff
@@ -189,3 +189,6 @@ apps/stable_diffusion/web/models/
 
 # Stencil annotators.
 stencil_annotator/
+
+# brevitas custom op lib
+apps/language_models/scripts/vicuna.py
```
Abhishek-Varma, line 194: This doesn't seem correct. Can you confirm?

Abhishek-Varma (commented):
Please address one comment - rest looks good.

.gitignore
powderluv: if you add it here it wont be detected by git. You want to add it to the black format command line with a `--exclude` or something

apps/language_models/scripts/vicuna.py
```diff
@@ -58,12 +90,1408 @@
...
+            self.first_vicuna_vmfb_path = self.get_model_path(suffix="vmfb")
+        if self.second_vicuna_vmfb_path == None:
+            self.second_vicuna_vmfb_path = self.get_model_path(
+                "second", "vmfb"
+            )
+        self.tokenizer = self.get_tokenizer()
+        self.shark_model = self.compile()
+
+    def get_model_path(self, model_number="first", suffix="mlir"):
+        safe_device = "_".join(self.device.split("-"))
```
Abhishek-Varma, line 904: Replace this with `safe_device = self.device.split("-")[0]`

apps/language_models/scripts/vicuna.py
```diff
@@ -58,12 +90,1408 @@
...
+                )
+        else:
+            # get second vic
+            # TODO: Remove after testing to avoid memory overload
+            # svic_shark_model = self.compile_second_vicuna()
+            pass
+
+        return None
+        # return tuple of shark_modules once mem is supported
+        # return fvic_shark_model, svic_shark_model
```
Abhishek-Varma, line 1323: Replace this with :-
```
def compile(self):
        # Cannot load both the models in the memory at once
        # due to memory constraints, hence on demand compilation
        # is being used until the space is enough for both models

        # Testing : DO NOT Download Vmfbs if not found. Modify later
        # download vmfbs for A100
        supported_devices = ["cuda", "cpu-sync", "cpu-task", "cpu"]
        if (
            not self.first_vicuna_vmfb_path.exists()
            and self.device in supported_devices
            and self.precision in ["fp32", "fp16", "int8"]
        ):
            if (self.device == "cuda" and self.precision == "fp16") or (
                self.device in ["cpu-sync", "cpu-task"]
                and self.precision == "int8"
            ):
                download_public_file(
                    f"gs://shark_tank/vicuna/unsharded/vmfb/{self.first_vicuna_vmfb_path.name}",
                    self.first_vicuna_vmfb_path.absolute(),
                    single_file=True,
                )
            else:
                pass

        else:
            # get first vic
            # TODO: Remove after testing to avoid memory overload
            # fvic_shark_model = self.compile_first_vicuna()
            pass
        if (
            not self.second_vicuna_vmfb_path.exists()
            and self.device in supported_devices
            and self.precision in ["fp32", "fp16", "int8"]
        ):
            if (self.device == "cuda" and self.precision == "fp16") or (
                self.device in ["cpu-sync", "cpu-task"]
                and self.precision == "int8"
            ):
                download_public_file(
                    f"gs://shark_tank/vicuna/unsharded/vmfb/{self.second_vicuna_vmfb_path.name}",
                    self.second_vicuna_vmfb_path.absolute(),
                    single_file=True,
                )
            else:
                pass
        else:
            # get second vic
            # TODO: Remove after testing to avoid memory overload
            # svic_shark_model = self.compile_second_vicuna()
            pass

        return None
        # return tuple of shark_modules once mem is supported
        # return fvic_shark_model, svic_shark_model
```

Abhishek-Varma (commented):
Please address the following sub-comments and also make changes to the following file in order to integrate with WebUI as well :-

Replace the set of lines at : https://github.com/nod-ai/SHARK/blob/91ab594744ffbe982ababe340fcd208923ecae48/apps/stable_diffusion/web/ui/stablelm_ui.py#L45-L47

With : 
```
from apps.language_models.scripts.vicuna import (
            UnshardedVicuna,
)            
```

Replace the following : https://github.com/nod-ai/SHARK/blob/91ab594744ffbe982ababe340fcd208923ecae48/apps/stable_diffusion/web/ui/stablelm_ui.py#L61

With :
```
vicuna_model = UnshardedVicuna(
```

PR #1607, opened 2023-06-28
Update README.md

karan-kankariya:
Cleaned up installation guides

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,129 +1,114 @@
 # SHARK
 
-High Performance Machine Learning Distribution
+ SHARK is a high performarce machine learning distribution that is built on LLVM, MLIR, OpenXLA’s IREE and nod.ai’s tuning.
 
 [![Nightly Release](https://github.com/nod-ai/SHARK/actions/workflows/nightly.yml/badge.svg)](https://github.com/nod-ai/SHARK/actions/workflows/nightly.yml)
 [![Validate torch-models on Shark Runtime](https://github.com/nod-ai/SHARK/actions/workflows/test-models.yml/badge.svg)](https://github.com/nod-ai/SHARK/actions/workflows/test-models.yml)
 
-
-<details>
-  <summary>Prerequisites - Drivers </summary>
-  
-#### Install your Windows hardware drivers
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,129 +1,114 @@
 # SHARK
 
-High Performance Machine Learning Distribution
+ SHARK is a high performarce machine learning distribution that is built on LLVM, MLIR, OpenXLA’s IREE and nod.ai’s tuning.
```
monorimet, line 3: typo

monorimet (changes requested):
Looks good, there are a few parts that could use updating but this reads better and doesn't hide macos/linux instructions which is good.

README.md
```diff
@@ -1,129 +1,114 @@
...
 ```
+<img width="1607" alt="webui" src="https://user-images.githubusercontent.com/74956/204939260-b8308bc2-8dc4-47f6-9ac0-f60b66edab99.png">
+
+## MacOS & Linux Guide
 
-#### Linux / macOS Users
+### Prerequisites
+* Install Python 3.11
+#### MacOS Drivers
+* Install the 1.3.216 Vulkan SDK from [here](https://sdk.lunarg.com/sdk/download/1.3.216.0/mac/vulkansdk-macos-1.3.216.0.dmg). Note: Newer versions of the SDK will not work.
```
powderluv, line 85: We no longer need this. We have the Metal backend.

cqtqlyst:
would adding hardware prerequisites especially for windows & linux be useful?

README.md
```diff
@@ -1,129 +1,114 @@
...
+## Windows 10/11 QuickStart Guide
 
-<details>
-  <summary>Prerequisites - Drivers </summary>
-  
-#### Install your Windows hardware drivers
+### Installation
+
+Install Windows Hardware Drivers
 * [AMD RDNA Users] Download the latest driver [here](https://www.amd.com/en/support/kb/release-notes/rn-rad-win-23-2-1).
```
one-lithe-rune, line 14: Now that MLIR/IREE support is integrated into the mainline drivers, I think its a bit problematic to link to a particular driver version as the 'latest', as we're not set up to keep updating to link to be to the actual latest one.

Maybe we should just say > 23.2.1 and link to the main AMD support page at https://www.amd.com/en/support

PR #1610, opened 2023-06-29
Update precision check for vicuna

jinchen62:


dan-garvey (approved):
lgtm but long term we should add compilation support here

PR #1611, opened 2023-06-30
changed method of compiling vicuna to remove first and second vicuna

Eliasj42:


apps/language_models/src/pipelines/vicuna_sharded_pipeline.py
```diff
@@ -193,6 +311,7 @@ def compile_lmhead(
             device=device,
             mlir_dialect="tm_tensor",
             device_idx=device_idx,
+            mmap=False,
```
powderluv, line 314: shouldn't this be true ?

powderluv (approved):

PR #1614, opened 2023-07-01
adding path to correct pytorch binary

Ranvirsv:


apps/stable_diffusion/shark_sd.spec
```diff
@@ -42,7 +42,12 @@ datas += [
          ( 'web/ui/logos/*', 'logos' )
          ]
 
+torch_path = sys.prefix + '/lib/python3.11/site-packages/torch/lib/libtorch_python.dylib'
```
powderluv, line 45: This won't work on Windows. I think you can just add to copy_data()

powderluv (changes requested):

apps/stable_diffusion/shark_sd.spec
Ranvirsv: the spec file already has copy_metadata('torch')

Ranvirsv:
Still has some issues with iree-compile.

PR #1615, opened 2023-07-01
Add the shark compile downstream due to https://github.com/pytorch/py…

pashu123:
…torch/pull/104185#issuecomment-1615110613

The registration can be done downstream and can be maintained there.

### DIFF
--- a/shark/__init__.py
+++ b/shark/__init__.py
@@ -1,0 +1,28 @@
+import importlib
+import logging
+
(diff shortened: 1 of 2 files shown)


pashu123:
@powderluv Due to https://github.com/pytorch/pytorch/pull/104185#issuecomment-1615110613 I have added the registration downstream.

powderluv (approved):

PR #1618, opened 2023-07-04
minor fix, some changes, some additions, and cleaning up

xzuyn:
- fix overflowing text (a janky fix).
- increase prompt and negative prompt input box sizes to 2 lines so long default prompts don't get cut off.
- set window size to 81% monitor width and 91% monitor height since it fits the screen better than doing a non-maximized window at the monitor's full resolution. would be better to just maximize the window on launch, but I can't figure that out yet.
- add `DEISMultistep`, `DPMSolverSingleStep`, `KDPM2AncestralDiscrete`, & `HeunDiscrete` schedulers as an option.
- set default prompt to something different/newer (maybe we could change it every stable release, or every couple stable releases to spice things up?).
- set default negative prompt to something different/newer.
- set default CFG to 3.5 (may be a little low for some people, but I find 3.5-5 works best).
- set default steps to 32. looks good at this amount or even as low as 16. will be a nice generation time boost for people sticking to default values.
- add `xzuyn/PhotoMerge` as a model option (this is my model, but i can remove it if needed).
- add `xzuyn/PhotoMerge-inpainting` as a model option.
- add 3 new example prompts.
- cleaning up and formatting.

Comparison of DEIS against existing schedulers.
sharkeulerdiscrete_deismultistep_dpmsolvermultistep

DEIS CFG Scale comparison.
3 5cfg vs 5cfg vs 7cfg

DEIS Step Count comparison.
8steps vs 16steps vs 32steps vs 64steps

### DIFF
--- a/apps/stable_diffusion/web/ui/img2img_ui.py
+++ b/apps/stable_diffusion/web/ui/img2img_ui.py
@@ -105,7 +105,7 @@ def img2img_inf(
             return (
                 None,
                 "Please provide either custom model or huggingface model ID, "
-                "both must not be empty",
+                "both must not be empty.",
             )
         if "civitai" in hf_model_id:
             args.ckpt_loc = hf_model_id
--- a/apps/stable_diffusion/web/ui/img2img_ui.py
+++ b/apps/stable_diffusion/web/ui/img2img_ui.py
@@ -326,7 +326,7 @@ def img2img_api(
         InputData["seed"],
         batch_count=1,
         batch_size=1,
-        scheduler="DEISMultistep",
+        scheduler="EulerDiscrete",
         custom_model="None",
         hf_model_id=InputData["hf_model_id"]
         if "hf_model_id" in InputData.keys()
@@ -520,7 +520,7 @@ with gr.Blocks(title="Image-to-Image") as img2img_web:
                         scheduler = gr.Dropdown(
                             elem_id="scheduler",
                             label="Scheduler",
-                            value="DEISMultistep",
+                            value="EulerDiscrete",
                             choices=scheduler_list_cpu_only,
                         )
                         with gr.Group():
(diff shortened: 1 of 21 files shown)


powderluv:
does SD 1.4 still work ok for you ? https://github.com/nod-ai/SHARK/actions/runs/5450064860/jobs/9914950812?pr=1618

xzuyn:
> does SD 1.4 still work ok for you ? https://github.com/nod-ai/SHARK/actions/runs/5450064860/jobs/9914950812?pr=1618

`CompVis/stable-diffusion-v1-4` works.

```
prompt=['a photo taken of the front of a super-car drifting on a road near mountains at high speeds with smokes coming off the tires, front angle, front point of view, trees in the mountains of the background, ((sharp focus))']
negative prompt=['watermark, signature, logo, text, lowres, ((monochrome, grayscale)), blurry, ugly, blur, oversaturated, cropped,']
model_id=CompVis/stable-diffusion-v1-4, ckpt_loc=
scheduler=DEISMultistep, device=Radeon (TM) RX 480 Graphics => vulkan://0
steps=32, guidance_scale=3.5, seed=[2934956637]
size=384x512, batch_count=1, batch_size=1, max_length=64

Clip Inference time (ms) = 70.999
Average step time: 791.3454622030258ms/it
VAE Inference time (ms): 819.717
Total image(s) generation time: 26.4309sec
```

apps/stable_diffusion/web/ui/img2img_ui.py
```diff
@@ -323,7 +326,7 @@ def img2img_api(
         InputData["seed"],
         batch_count=1,
         batch_size=1,
-        scheduler="EulerDiscrete",
+        scheduler="DEISMultistep",
```
powderluv, line 329: Can we leave this as EulerDiscrete for now ? We will need to update all the golden images otherwise. We can do that separately if we want.

xzuyn: just img2img, or all the ones I changed to DEIS by default?

powderluv: all of them since I think we have golden tests for each. We can land a default scheduler change separately and update the goldens in one PR. We could also offload it to GPU if we want.

powderluv:
@dan-garvey can we please update the golden images after this patch lands ?

PR #1620, opened 2023-07-05
int8 e2e for WebUI

Abhishek-Varma:


vivekkhandelwal1 (approved):
LGTM

apps/language_models/scripts/vicuna.py
```diff
@@ -103,7 +103,14 @@
 )
 
 
-def brevitas〇matmul_rhs_group_quant〡shape(lhs: List[int], rhs: List[int], rhs_scale: List[int], rhs_zero_point: List[int], rhs_bit_width: int, rhs_group_size: int) -> List[int]:
+def brevitas〇matmul_rhs_group_quant〡shape(
```
jinchen62, line 106: Wouldn't you have the signature mismatch error with this change?

powderluv: Also we should remove the unicode `〇`

PR #1624, opened 2023-07-06
Xfail models for data tiling flag changes

monorimet:


powderluv:
im trying out a change that was suggested in another PR. If that doesn't work we can land this.

PR #1626, opened 2023-07-06
[Vicuna] Revert the formatting for Brevitas op

Abhishek-Varma:
-- This commit reverts the formatting for Brevitas op.
-- It also excludes vicuna.py script from `black` formatter.

### DIFF
--- a/apps/language_models/scripts/vicuna.py
+++ b/apps/language_models/scripts/vicuna.py
@@ -103,14 +103,7 @@ parser.add_argument(
 )
 
 
-def brevitas〇matmul_rhs_group_quant〡shape(
-    lhs: List[int],
-    rhs: List[int],
-    rhs_scale: List[int],
-    rhs_zero_point: List[int],
-    rhs_bit_width: int,
-    rhs_group_size: int,
-) -> List[int]:
+def brevitas〇matmul_rhs_group_quant〡shape(lhs: List[int], rhs: List[int], rhs_scale: List[int], rhs_zero_point: List[int], rhs_bit_width: int, rhs_group_size: int) -> List[int]:
     if len(lhs) == 3 and len(rhs) == 2:
         return [lhs[0], lhs[1], rhs[0]]
     elif len(lhs) == 2 and len(rhs) == 2:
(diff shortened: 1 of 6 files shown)


apps/language_models/scripts/vicuna.py
```diff
@@ -103,14 +103,7 @@
...
 
-def brevitas〇matmul_rhs_group_quant〡shape(
-    lhs: List[int],
-    rhs: List[int],
-    rhs_scale: List[int],
-    rhs_zero_point: List[int],
-    rhs_bit_width: int,
-    rhs_group_size: int,
-) -> List[int]:
+def brevitas〇matmul_rhs_group_quant〡shape(lhs: List[int], rhs: List[int], rhs_scale: List[int], rhs_zero_point: List[int], rhs_bit_width: int, rhs_group_size: int) -> List[int]:
```
powderluv, line 106: @stellaraccident  do you know how this changes the function signature.

powderluv (approved):

powderluv:
@Abhishek-Varma  but why revert the other files

powderluv (changes requested):
Why the changes the pyproject too?

Abhishek-Varma:
Hi @powderluv 

I wasn't including the other files in this patch initially but the CI showed linting failures in those set of files - opt_params and model_wrappers.py.
So I applied `black` formatting on those as well and added them.

Also, pyproject was changed to exclude `vicuna.py` from being formatted by black - this is with reference to [Jinchen's comment here](https://github.com/nod-ai/SHARK/pull/1598#discussion_r1251617853) and also because of the formatting it doesn't get registered and fails the lowering at torch-mlir level itself. CC: @vivekkhandelwal1

apps/language_models/scripts/vicuna.py
powderluv: And @makslevental  any thoughts

powderluv:
Opened #1627 to track this issue. so we can unblock the demos

apps/language_models/scripts/vicuna.py
vivekkhandelwal1: I think this is because of the way these functions are verified in Torch-MLIR. They match the functions' first line for verifying the signature (https://github.com/llvm/torch-mlir/blob/main/python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/library_generator.py#L177), and after formatting the function definition goes to multiple lines and as a result, it would fail the verification.

PR #1632, opened 2023-07-07
Fix pytest benchmarks and shark_tank generation.

monorimet:
- fix setup_venv.sh for benchmarks/imports etc.
- unpin torch-mlir
- fix torch benchmarks in SharkBenchmarkRunner
- generate SD artifacts using `build_tools/stable_diffusion_testing.py` and `--import_mlir`
- decouple SD gen from `tank/generate_sharktank` for now


monorimet:
https://github.com/openxla/iree/pull/14303 updates some experimental TF `mlir` API imports but with current pypi versions of tf-nightly the API does not match. Switching tf model generation off for now.

setup.py
```diff
@@ -39,7 +39,7 @@
     install_requires=[
         "numpy",
         "PyYAML",
-        "torch-mlir",
+        "torch-mlir==20230802.121",
```
powderluv, line 42: wait wait why

monorimet: Just testing, it wouldn't pull the newest and was instead pulling from earlier today.

monorimet: We don't need the pin when merging

.github/workflows/test-models.yml
```diff
@@ -157,7 +157,7 @@ jobs:
     - name: Validate Stable Diffusion Models (Windows)
       if: matrix.suite == 'vulkan' && matrix.os == '7950x'
       run: |
-        ./setup_venv.ps1
+        ./setup_venv.ps1 --force
```
powderluv, line 160: shouldn't need this. This should be a new venv ?

setup.py
monorimet: Resolving as this pin was removed.

.github/workflows/test-models.yml
monorimet: Makes sense -- I think I was trying to get the right torch-mlir version but it was a different issue.
Removed this change.

shark/shark_benchmark_runner.py
```diff
@@ -124,42 +124,48 @@ def benchmark_frontend(self, modelname):
...
+            frontend_model.cpu()
+            input.cpu()
         # TODO: re-enable as soon as pytorch CUDA context issues are resolved
-        try:
-            frontend_model = torch.compile(
-                frontend_model, mode="max-autotune", backend="inductor"
-            )
-        except RuntimeError:
-            frontend_model = HFmodel.model
+        # try:
```
dan-garvey, line 152: dead code?

dan-garvey (dismissed):

shark/shark_benchmark_runner.py
monorimet: Yeah even if we enable this again it probably will need updating -- getting GPU baselines with inductor in these pytests is quite a mess.

powderluv (approved):

PR #1634, opened 2023-07-07
UI/Web Output gallery - add 'open folder' button (needs Mac/Linux testing)

one-lithe-rune:
_This contains code for Mac and Linux which I have not tested so I don't know is correct. Plus the whole concept feels pretty half-assed to me, but then again I keep running from the branch I have it on and **using** the thing. And continually rebasing it. Soooo PR._ 

### Motivation 

It would be useful to be able to mark images in the output gallery and then filter them in or out of the gallery. Also at some stage move the ones you don't want somewhere else so they can be deleted (or move the ones you want to keep somewhere else, so you can delete what's left). I was trying to think of the simplest ways to do that whilst having the filtering persist between runs, and one of the ways I thought of was 'having marking an image move it to another subdirectory, and unmarking it move it back'.

This is not that.

This is 'Well if you're thinking of faffing about shuffling files around like that, you might as well just throw up your hands and put in a ~expletive~ button that opens the current subdir in the os file browser if you're running the UI on the same machine Shark is running, have them do it manually, and expect them to hit refresh we they're done. Wait! is that even possible?'

Apparently yes.

### Changes

* Adds a button that opens the currently selected subdirectory using the default OS file manager if the http request is from 127.0.0.1 or localhost.
* Improves output gallery handling of having images deleted out from under it.
* Don't show VAE or LoRA lines in output gallery parameter info panel when their value is 'None'

### Problems/Concerns

* I've written code to do this for Windows, Mac and Linux. I don't know that what I've written works at all on Mac or Linux. So probably DNM until someone who can test those has done so.
* In Windows the first time you use in a session it the File Explorer flashes discretely to let you know it wants attention rather than immediately popping up the window.
* Entirely useless if you're not on the same machine as the the browser.
* Doesn't really solve the original filtering problem, because you still have to find the files you want muck about with in os file browser even tho' it is pointing at the right directory, and then you have to refresh once you're done, and... 🤷

### DIFF
--- a/apps/stable_diffusion/web/ui/css/sd_dark_theme.css
+++ b/apps/stable_diffusion/web/ui/css/sd_dark_theme.css
@@ -247,7 +247,7 @@ footer {
     line-height: var(--line-xs)
 }
 
-#output_refresh_button {
+.output_icon_button {
     max-width: 30px;
     align-self: end;
     padding-bottom: 8px;
--- a/apps/stable_diffusion/web/ui/outputgallery_ui.py
+++ b/apps/stable_diffusion/web/ui/outputgallery_ui.py
@@ -1,6 +1,8 @@
 import glob
 import gradio as gr
 import os
+import subprocess
+import sys
 from PIL import Image
 
 from apps.stable_diffusion.src import args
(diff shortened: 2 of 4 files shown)


powderluv:
LMK when you rebase to latest gradio and I can test on mac.

one-lithe-rune:
Conflict should now be fixed.

powderluv (approved):

PR #1637, opened 2023-07-08
UI/Web: Some Fixes for Gradio 3.36.1

one-lithe-rune:
### Motivation

Since Gradio is now unpinned I'm getting Gradio 3.36.1 as the latest, which has broken some stuff cosmetically and caused a lot of warnings on startup due to the deprecation of `.style` on controls.

### Changes

* Clear .style deprecation warnings.
* Re-remove download button from Nod logos.
* Add work around for `container=false` not doing what it did before on dropdowns to the output gallery CSS

### Problems/Concerns

* I've not linted with black prior to check-in as for some reason `block --check .` locally has it suddenly deciding to lint my entire `shark.venv` when I do that and I have kill as it hits 100% CPU.
* Only works around the `container=False` problem for the output gallery. This setting was also used on the Chatbot and (I think!) other places, so those get whatever the gradio current styling for that case.

### DIFF
--- a/apps/stable_diffusion/web/ui/css/sd_dark_theme.css
+++ b/apps/stable_diffusion/web/ui/css/sd_dark_theme.css
@@ -227,8 +227,17 @@ footer {
 }
 
 /* Hide the download icon from the nod logo */
-#top_logo .download {
+#top_logo button {
     display: none;
+}
+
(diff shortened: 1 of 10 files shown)


powderluv:
thank you. btw you can cd into a dir and then run `black .` in the future.

PR #1639, opened 2023-07-09
Added support for building ZIP distributions

AyaanShah2204:
- Run `pyinstaller .\apps\stable_diffusion\studio_bundle.spec -y` to build a single-directory application for SHARK
- Run `python .\build_tools\bundled_app_to_zip.py` to build a zip file in the main directory

onefile exe applications are still created in the same way they were before: `pyinstaller .\apps\stable_diffusion\shark_sd.spec`

Reason:
`--onefile` and `--onedir` are commands that pyinstaller accepts when it is pointed to a python script. These commands dictate how a spec file is created. Once the spec file is created, it will produce either a folder or an exe file depending on how the spec file was configured
Hence, if we want our build commands to point to spec files (because it allows us to use hidden imports and pyinstaller hooks without writing an unwieldy shell command) we need separate spec files for exe and zip

### DIFF
--- a/apps/stable_diffusion/studio_bundle.spec
+++ b/apps/stable_diffusion/studio_bundle.spec
@@ -1,59 +1,9 @@
 # -*- mode: python ; coding: utf-8 -*-
-from PyInstaller.utils.hooks import collect_data_files
-from PyInstaller.utils.hooks import copy_metadata
-from PyInstaller.utils.hooks import collect_submodules
-
-import sys ; sys.setrecursionlimit(sys.getrecursionlimit() * 5)
-
-datas = []
-datas += collect_data_files('torch')
-datas += copy_metadata('torch')
-datas += copy_metadata('tqdm')
-datas += copy_metadata('regex')
-datas += copy_metadata('requests')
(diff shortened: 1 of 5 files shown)


build_tools/bundled_app_to_zip.py
powderluv, line 1: lets drop this file. We can run this from the command line.

apps/stable_diffusion/studio_bundle.spec
powderluv, line 1: can we move all of the common parts of both .spec files to a shark_studio_imports.py and then lets import into both .spec files.

build_tools/bundled_app_to_zip.py
AyaanShah2204: Done! We can use `Compress-Archive` to zip the bundle from nightly build scripts.

apps/stable_diffusion/studio_bundle.spec
AyaanShah2204: Moved all the heavy looking stuff :)

powderluv (approved):

Issue #1351, opened 2023-04-25
Batch Seeds Improvement


nirvedhmeshram:

one-lithe-rune:
Probably not what you're talking here but...

...maybe have an option for the first seed in a batch to be used to seed the random number generator that gets the rest of the seeds for the batch? That would mean if you used the same first seed with that option you would get the same seeds for the rest of the batch every time, making batches repeatable with respect to seeds.

PR #1654, opened 2023-07-15
SD - Add repeatable (batch) seeds option

one-lithe-rune:
Partially addresses https://github.com/nod-ai/SHARK/issues/1351

### Motivation

When dialing in prompts I often want to see multiple images with my prompt changes, but also compare them to images generated with a previous prompt, currently this requires you to generate an single image at a time, entering each seed one by one.

These changes allow you to set batches to use the same set of seeds as a previous set of batches when the first batch has the same starting seed. You can then re-run those batches with each batch getting the same seed as before.

### Changes

* Generates the seeds for all batch_count batches being run up front rather than generating the seed for a batch just before it is run.
* Adds a --repeatable_seeds argument defaulting to False
* When repeatable_seeds=True, the first seed for a set of batches will also be used as the rng seed to generate the subsequent batch seeds in the run. The rng seed is reset once all seeds are generated.
* When repeatable_seeds=False, batch seeding works as currently.
* Updates scripts under apps/scripts that support the batch_count argument to also support the repeatable_seeds argument.
* UI/Web: Adds a checkbox element on each SD tab after batch count/size for toggling repeatable seeds, and update _inf functions to take this into account.
* UI/Web: Moves the Stop buttons out of the Advanced sections and next to Generate to make things not fit quite so badly with the extra UI elements.
* UI/Web: Fixes logging to the upscaler output text box not working correctly when running multiple batches.

* It would be unsurprising if I missed somewhere where batches are used.
* I haven't tested the changes to the scripts, though I have updated them.
* I haven't tested the API in light of the changes, though I have updated the functions.
* If batch_size > 1 this probably won't have the desired effect regardless of batch_count.

### DIFF
--- a/apps/stable_diffusion/scripts/inpaint.py
+++ b/apps/stable_diffusion/scripts/inpaint.py
@@ -58,11 +58,8 @@ def main():
         ondemand=args.ondemand,
     )
 
+    seeds = utils.batch_seeds(seed, args.batch_count, args.repeatable_seeds)
     for current_batch in range(args.batch_count):
-        if current_batch > 0:
-            seed = -1
-        seed = utils.sanitize_seed(seed)
-
         start_time = time.time()
         generated_imgs = inpaint_obj.generate_images(
             args.prompts,
@@ -76,7 +73,7 @@ def main():
             args.inpaint_full_res_padding,
             args.steps,
             args.guidance_scale,
-            seed,
+            seeds[current_batch],
             args.max_length,
             dtype,
             args.use_base_vae,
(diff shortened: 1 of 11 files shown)


powderluv:
Only comment is about `UI/Web: Moves the Stop buttons out of the Advanced sections and next to Generate to make things not fit quite so badly with the extra UI elements.`

I don't think the Stop button works when you start generating the first image, or maybe it is fixed. Anyway we want to move the compile steps into the model manager so hopefully it is easy to fixup stop.

one-lithe-rune:
I would be surprised if the stop batch button worked anywhere but between images in a batch (or between tiles for upscaler), as I don't think the object that holds the cancel flag gets passed down far enough into the pipeline for it even known that you've pressed it until you get back into the UI code.

powderluv:
Yeah that is why we hid it under advanced for now

PR #1667, opened 2023-07-18
set optional vmfb download

dan-garvey:


apps/language_models/scripts/vicuna.py
```diff
@@ -1322,6 +1325,7 @@ def compile(self):
             if (self.device == "cuda" and self.precision == "fp16") or (
                 self.device in ["cpu-sync", "cpu-task"]
                 and self.precision == "int8"
+                and self.download_vmfb
```
jinchen62, line 1328: Could you make this and line 1305 in same format though there is no lint for this file.

jinchen62 (approved):

PR #1673, opened 2023-07-19
Add Llama2 70B option in CLI and WebUI

Abhishek-Varma:


powderluv (approved):

vivekkhandelwal1:
I have to revert this patch, since this results in an issue with the existing llama-2-7b.

PR #1675, opened 2023-07-19
Add pyyaml metadata to shark studio .exe imports

monorimet:


powderluv:
sorry just saw this PR but landed the latter one by @AyaanShah2204

PR #1681, opened 2023-07-20
Add Langchain CPU support and update requirements

vivekkhandelwal1:


powderluv (approved):

powderluv:
We should migrate those requirements to our main one

vivekkhandelwal1:
> We should migrate those requirements to our main one

Will do that gradually, once we are confirmed with what all out of these are actually required.

PR #1694, opened 2023-07-26
Langchain fix for CPU, remove max time stopping criteria

vivekkhandelwal1:


powderluv:
This does seem like a legit failure

vivekkhandelwal1:
> This does seem like a legit failure

But it's failing the stable diffusion test, I don't understand how my patch affects those tests.

@dan-garvey @monorimet any idea?

PR #1697, opened 2023-07-27
Fix and reformat vicuna script

jinchen62:
and some cleanup


dan-garvey (dismissed):
approved assuming you tested the configurations

dan-garvey (approved):

PR #1699, opened 2023-07-28
Change the DocuChat web UI

gpetters94:
This adds an upload tab for DocuChat. With this the only remaining UI work is to plumb in the precision selector.

### DIFF
--- a/apps/stable_diffusion/web/index.py
+++ b/apps/stable_diffusion/web/index.py
@@ -115,6 +115,7 @@ if __name__ == "__main__":
         txt2img_sendto_inpaint,
         txt2img_sendto_outpaint,
         txt2img_sendto_upscaler,
+        h2ogpt_upload,
         h2ogpt_web,
         img2img_web,
         img2img_custom_model,
@@ -247,7 +248,9 @@ if __name__ == "__main__":
                         upscaler_status,
                     ]
                 )
-            with gr.TabItem(label="DocuChat(Experimental)", id=11):
+            with gr.TabItem(label="DocuChat Upload", id=11):
+                h2ogpt_upload.render()
+            with gr.TabItem(label="DocuChat(Experimental)", id=12):
                 h2ogpt_web.render()
 
         # send to buttons
--- a/apps/stable_diffusion/web/ui/h2ogpt.py
+++ b/apps/stable_diffusion/web/ui/h2ogpt.py
@@ -12,6 +12,10 @@ from apps.language_models.langchain.enums import (
     LangChainAction,
 )
 import apps.language_models.langchain.gen as gen
+from gpt_langchain import (
+    path_to_docs,
+    create_or_update_db,
+)
 from apps.stable_diffusion.src import args
 
 
@@ -168,7 +172,14 @@ def chat(curr_system_message, history, device, precision):
     return history
 
 
-with gr.Blocks(title="H2OGPT") as h2ogpt_web:
+userpath_selector = gr.Textbox(
+    label="Document Directory",
+    value=str(os.path.abspath("apps/language_models/langchain/user_path/")),
+    interactive=True,
+    container=True,
+)
+
+with gr.Blocks(title="DocuChat") as h2ogpt_web:
     with gr.Row():
         supported_devices = available_devices
         enabled = len(supported_devices) > 0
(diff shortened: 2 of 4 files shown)


apps/stable_diffusion/web/index.py
```diff
@@ -244,7 +245,9 @@ def register_outputgallery_button(button, selectedid, inputs, outputs):
                         upscaler_status,
                     ]
                 )
-            with gr.TabItem(label="DocuChat(Experimental)", id=10):
+            with gr.TabItem(label="DocuChat Upload", id=10):
+                h2ogpt_upload.render()
```
vivekkhandelwal1, line 249: If I'm not wrong this just adds a tab for uploading the documents, right? This doesn't create the vector database.

gpetters94: I thought the database was built during the call to `evaluate`, but if not I can hop on for a few hours on Saturday and add that.

apps/language_models/langchain/gpt_langchain.py
```diff
@@ -372,7 +372,7 @@ def get_embedding(
         from langchain.embeddings import HuggingFaceEmbeddings
 
         torch_dtype, context_class = get_dtype()
-        model_kwargs = dict(device=args.device)
+        model_kwargs = dict(device="cpu")
```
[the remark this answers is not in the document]
vivekkhandelwal1: @gpetters94, I don't see any reason for hardcoding the device CPU here.

apps/stable_diffusion/web/ui/h2ogpt.py
```diff
@@ -168,10 +172,106 @@ def chat(curr_system_message, history, device, precision):
...
+        with gr.Row():
+            upload = gr.UploadButton(
+                label="Upload documents",
+                file_count="multiple",
+            )
+            upload.upload(fn=upload_file, inputs=upload, outputs=text)
+            userpath_selector.render()
+            userpath_selector.input(
+                fn=update_userpath, inputs=userpath_selector, outputs=text
+            ).then(fn=update_or_create_db)
```
vivekkhandelwal1, line 271: I think you should add this code to a different file. Because on using both the features in the same session, the database creation happens to be running twice.

vivekkhandelwal1 (changes requested):
@gpetters94, if the document is uploaded from the web UI it takes a large amount of time for database creation as compared to manually placing it in the user_path folder directly. For example, for a doc it takes nearly 7 minutes from web UI and few seconds from the CLI. You need to take a look at this issue as well.

apps/stable_diffusion/web/ui/h2ogpt.py
gpetters94: I don't see it running twice. If you mean that it runs on every file upload and once at the start that's expected behavior (in case the user populates the directory before the web UI is started).

vivekkhandelwal1: Could you also verify the amount of time both ways? With a document sized in MBs?

gpetters94: After adding those files and starting up it takes 110 seconds. I've also verified that it's only running once per upload and once at the start.

vivekkhandelwal1: Just a small question, why do I see this
```
0it [00:00, ?it/s][ProgressParallel(n_jobs=-1)]: Using backend MultiprocessingBackend with 12 concurrent workers.
Ingesting file: /home/vivek/work/shark-vivekkhandelwal1/apps/language_models/langchain/user_path/./Abraham Silberschatz, Greg Gagne, Peter B. Galvin - Operating System Concepts-Wiley (2018).pdf
Ingesting file: /home/vivek/work/shark-vivekkhandelwal1/apps/language_models/langchain/user_path/./operating_system.txt
 50%|██████████████████████████████████████████████████████████████████████████████████████████████                                                                                              | 1/2 [00:00<00:00,  1.99it/s][ProgressParallel(n_jobs=-1)]: Batch computation too fast (0.02492046356201172s.) Setting batch_size=2.
100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 2/2 [00:22<00:00, 11.27s/it]
0it [00:00, ?it/s]
Loading and updating db
Found 0 new sources (0 have no hash in original source, so have to reprocess for migration to sources with hash)
Removing 0 duplicate files from db because ingesting those as new documents
```
Even before choosing/running any model?

vivekkhandelwal1: Shouldn't all this run when we select DocuChat? This way, those who don't want to run DocuChat would also have to go through this.

gpetters94: I see what you mean, though I imagine someone not running DocuChat wouldn't be populating the directory.

gpetters94: Okay, I've changed it to only run upon switching to either DocuChat tabs for the first time.

vivekkhandelwal1 (approved):

PR #1700, opened 2023-07-28
[Shard] Add sharding generation in shark studio

Shukla-Gaurav:


apps/stable_diffusion/web/index.py
```diff
@@ -225,6 +226,8 @@ def register_outputgallery_button(button, selectedid, inputs, outputs):
                 model_web.render()
             with gr.TabItem(label="Chat Bot(Experimental)", id=6):
                 stablelm_chat.render()
+            with gr.TabItem(label="Generate Sharding Config", id=6):
```
one-lithe-rune, line 229: Two tabs have `id=6` here.

one-lithe-rune: And having just accidentally ended up with two `id=9`, the effect of having duplicates appears to be the one of them disappears when you click on it.

PR #1702, opened 2023-07-29
Unsharded Vicuna: Fix Memory Error compiling mlir for lmsys/vicuna-7b-v1.3 fp16 with 64 GiB

one-lithe-rune:
### Motivation

I get a Memory Error writing the combined MLIR file inside the Python text encoding IO code. Presumably an OOM since it heavily hits the page file with my 64GiB of system memory.

This is on Windows 10, run through the UI on, with `lmsys/vicuna-7b-v1.3`, fp16, and vulkan selected.

These changes get me all the way through IREE compile which then falls over complaining about a constant redefinition.

### Changes

* in `combine_mlir_scripts` since we're writing a file anyway, read the file back in to get the return value, so we don't have to have another copy assembled in memory whilst writing the file.
* In `combine_mlir_scripts` write the file out line by line from the individual source vars instead of writing everything at once to reduce memory pressure when python does text encoding,
* In `combine_mlir_scripts` use del and gc.collect() more aggressively.
* Outside `combine_mlir_scripts` add more del and gc.collect statements in the unsharded vicuna compile code to try and reduce memory pressure for when we call that function.

### Problems Concerns

* Do we even care about unsharded these days?
* Probably slower if you have 128GiB of system memory, and its already slow.
* Some of the additional `del`s and `gc.collect`s probably aren't doing anything, they are all 'well that looks reasonable to me' placements, rather than having done any profiling.
* Haven't tried with the TheBloke/vicuna version of the model. Don't know whether it had this problem.

### DIFF
--- a/apps/language_models/scripts/vicuna.py
+++ b/apps/language_models/scripts/vicuna.py
@@ -1,6 +1,7 @@
 import argparse
 import json
 import re
+import gc
 from io import BytesIO
 from pathlib import Path
 from tqdm import tqdm
@@ -39,9 +40,6 @@ from shark.shark_inference import SharkInference
 
 from brevitas_examples.llm.llm_quant.quantize import quantize_model
 from brevitas_examples.llm.llm_quant.run_utils import get_model_impl
-
-if __name__ == "__main__":
-    import gc
 
 
 parser = argparse.ArgumentParser(
@@ -114,11 +112,18 @@ parser.add_argument(
     "--cache_vicunas",
     default=False,
     action=argparse.BooleanOptionalAction,
-    help="For debugging purposes, creates a first_{precision}.mlir and second_{precision}.mlir and stores on disk"
+    help="For debugging purposes, creates a first_{precision}.mlir and second_{precision}.mlir and stores on disk",
 )
 
 
-def brevitas〇matmul_rhs_group_quant〡shape(lhs: List[int], rhs: List[int], rhs_scale: List[int], rhs_zero_point: List[int], rhs_bit_width: int, rhs_group_size: int) -> List[int]:
+def brevitas〇matmul_rhs_group_quant〡shape(
+    lhs: List[int],
+    rhs: List[int],
+    rhs_scale: List[int],
+    rhs_zero_point: List[int],
+    rhs_bit_width: int,
+    rhs_group_size: int,
+) -> List[int]:
     if len(lhs) == 3 and len(rhs) == 2:
         return [lhs[0], lhs[1], rhs[0]]
     elif len(lhs) == 2 and len(rhs) == 2:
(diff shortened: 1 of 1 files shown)


one-lithe-rune:
There's more I think can do in this code to reduce memory usage, but it would involve shuffling around the statements that write the mlir, so that the file is written as we go along. 

Since I don't know whether unsharded is even a thing we care about given that sharded exists, and I don't have a output mlir file that is correct to compare against, I didn't think that was sensible to try that just to get me past the error.

powderluv:
Thank you for the PR. We do care about unsharded. Ideally on a single GPU system we should run unsharded. Right now we run sharded until we workaround a large memory allocation issue in Vulkan with Sparse Memory. That is being worked on by @benvanik upstream. 

@stellaraccident and @dan-garvey  probably are the best to chime in on the PR. But if you made it get to IREE compile without failing in a 64GB RAM system that is progress :D

dan-garvey:
Thanks for the PR. The failure at compile time is expected for fp16 currently. I will try to have a fix up for that in a few days.  Appreciate the help on the memory side!

apps/language_models/scripts/vicuna.py
```diff
@@ -983,7 +1024,7 @@ def __init__(
         low_device_memory=False,
         weight_group_size=128,
         download_vmfb=False,
-        cache_vicunas=False,
+        cache_vicunas=True,
```
dan-garvey, line 1027: Not sure it particularly matters here, but I'd prefer we leave this as false as it generates a lot of GB worth of files on disk

dan-garvey (dismissed):
One minor comment if you don't mind addressing

apps/language_models/scripts/vicuna.py
one-lithe-rune: Oops, thought I'd taken that out before committing. The command line flag wasn't working for me, so I hacked that in so I didn't have to generate the first and second vicuna data on every run.

one-lithe-rune: Should be fixed with https://github.com/nod-ai/SHARK/pull/1702/commits/0cb54b305a0e1af62c43185ba68aa90798c665be

Eliasj42 (dismissed):

apps/language_models/scripts/vicuna.py
```diff
@@ -114,11 +112,18 @@
     "--cache_vicunas",
     default=False,
     action=argparse.BooleanOptionalAction,
-    help="For debugging purposes, creates a first_{precision}.mlir and second_{precision}.mlir and stores on disk"
+    help="For debugging purposes, creates a first_{precision}.mlir and second_{precision}.mlir and stores on disk",
 )
 
 
-def brevitas〇matmul_rhs_group_quant〡shape(lhs: List[int], rhs: List[int], rhs_scale: List[int], rhs_zero_point: List[int], rhs_bit_width: int, rhs_group_size: int) -> List[int]:
```
jinchen62, line 121: pls revert the format

apps/language_models/scripts/vicuna.py
```diff
@@ -127,20 +132,30 @@ def brevitas〇matmul_rhs_group_quant〡shape(lhs: List[int], rhs: List[int], rh
         raise ValueError("Input shapes not supported.")
 
 
-def brevitas〇matmul_rhs_group_quant〡dtype(lhs_rank_dtype: Tuple[int, int], rhs_rank_dtype: Tuple[int, int], rhs_scale_rank_dtype: Tuple[int, int], rhs_zero_point_rank_dtype: Tuple[int, int], rhs_bit_width: int, rhs_group_size: int) -> int:
```
jinchen62, line 130: pls revert the format

apps/language_models/scripts/vicuna.py
```diff
@@ -127,20 +132,30 @@ def brevitas〇matmul_rhs_group_quant〡shape(lhs: List[int], rhs: List[int], rh
...
+    rhs_zero_point_rank_dtype: Tuple[int, int],
+    rhs_bit_width: int,
+    rhs_group_size: int,
+) -> int:
     # output dtype is the dtype of the lhs float input
     lhs_rank, lhs_dtype = lhs_rank_dtype
     return lhs_dtype
 
 
-def brevitas〇matmul_rhs_group_quant〡has_value_semantics(lhs, rhs, rhs_scale, rhs_zero_point, rhs_bit_width, rhs_group_size) -> None:
```
jinchen62, line 136: pls revert the format

apps/language_models/scripts/vicuna.py
```diff
@@ -127,20 +132,30 @@ def brevitas〇matmul_rhs_group_quant〡shape(lhs: List[int], rhs: List[int], rh
...
+def brevitas〇matmul_rhs_group_quant〡has_value_semantics(
+    lhs, rhs, rhs_scale, rhs_zero_point, rhs_bit_width, rhs_group_size
+) -> None:
     return
 
 
 brevitas_matmul_rhs_group_quant_library = [
     brevitas〇matmul_rhs_group_quant〡shape,
     brevitas〇matmul_rhs_group_quant〡dtype,
-    brevitas〇matmul_rhs_group_quant〡has_value_semantics]
```
jinchen62, line 143: pls revert the format

jinchen62 (changes requested):

apps/language_models/scripts/vicuna.py
one-lithe-rune: reverted in https://github.com/nod-ai/SHARK/pull/1702/commits/7c35d276c4a14e3a52de8c09b8a681cc86cc9777

apps/language_models/scripts/vicuna.py
one-lithe-rune: reverted in https://github.com/nod-ai/SHARK/pull/1702/commits/7c35d276c4a14e3a52de8c09b8a681cc86cc9777

apps/language_models/scripts/vicuna.py
one-lithe-rune: reverted in https://github.com/nod-ai/SHARK/pull/1702/commits/7c35d276c4a14e3a52de8c09b8a681cc86cc9777

jinchen62 (approved):

apps/language_models/scripts/vicuna.py
one-lithe-rune: Also not showing as outdated, so I haven't resolved these two. 

@jinchen62 please check if these are the format you need.

jinchen62: LGTM

PR #1708, opened 2023-08-02
[WIP] Windows Autoupdate Mechanism

AyaanShah2204:
For when we start distributing the Advanced Installer MSI:

- Shark Studio will run the updater.exe script to check and install upgrades
- This PR ensures that this only happens in Windows MSI installations

@powderluv should I keep the auto-update function in index.py or move it to a new file eg. `update.py` since we might have more update functions when we support Mac?

### DIFF
--- a/apps/stable_diffusion/web/index.py
+++ b/apps/stable_diffusion/web/index.py
@@ -40,23 +40,24 @@ def launch_app(address):
     )
     webview.start(private_mode=False)
 
+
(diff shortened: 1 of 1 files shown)


powderluv:
best to keep in a separate file.

AyaanShah2204:
Sounds good

powderluv:
this looks good but lets wait a day or so until the builds are good.

PR #1733, opened 2023-08-07
Clearly differentiate phases of loading modules to better understand if things are taking a long time.

stellaraccident:


monorimet (approved):
LGTM. The lint can be fixed with:
```
pip install black
black --line-length 79 .
```

PR #1734, opened 2023-08-07
Optimize device enumeration overhead and log details on long operations.

stellaraccident:
* Various fixes to add `@functools.cache` to what should be one time, expensive, device enumeration and setup activities. Cuts several seconds off of initialization on my machine.
* Add detailed tracing to actual invocations if they exceed a certain timeout or have an exception.
* Add detailed tracing to loading status.
* By default detail logging is only printed if an operation takes an excessive amount of time. All logging/timing can be printed by setting the variable `$env:SHARK_DETAIL_TRACE = "1"`

### DIFF
--- a/shark/iree_utils/_common.py
+++ b/shark/iree_utils/_common.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 ## Common utilities to be shared by iree utilities.
-
+import functools
 import os
 import sys
 import subprocess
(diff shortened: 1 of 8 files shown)


powderluv (dismissed):

monorimet (approved):
Thanks for this. LGTM

dan-garvey (dismissed):
Thank you for this

powderluv:
gah mac not too happy

PR #1738, opened 2023-08-08
Disable tunings on VAE by default and add --use_tuned_vae flag.

monorimet:


powderluv:
Lets hold on this. I think we have made a recreate without this.

PR #1739, opened 2023-08-09
re-enable constant folding

powderluv:
Tested and works well. (modulo unrelated driver issue)


monorimet (approved):

powderluv:
safetensors issue is an upstream issue

PR #1743, opened 2023-08-09
Remove accelerate from langchain

gpetters94:
This removes accelerate from langchain. It doesn't remove it from the main requirements.txt, since that might break things.


vivekkhandelwal1:
@gpetters94 is it fine, if I review this PR tomorrow after testing it locally?

PR #1754, opened 2023-08-13
add support passing iree flags for  LLMs

PhaneeshB:
- required for passing target triple from command line when running LLMs on web and cli
- NOTE:
  - target triple for Sharded LLMs is experimental as of now.
  - duplicate args from SD and LLMs will need to be refactored into a more disjoint set of cli args to prevent the argparse objects from eating up each others values when used in the same file.


jinchen62:
Does it work for cli?

PhaneeshB:
> Does it work for cli?

Yes @jinchen62 added support for cli too.
Thanks

jinchen62 (approved):
LGTM

PR #1755, opened 2023-08-13
Set default allocator for metal device creation

PhaneeshB:


shark/iree_utils/compile_utils.py
```diff
@@ -547,9 +547,15 @@ def get_results(
 def get_iree_runtime_config(device):
     device = iree_device_map(device)
     haldriver = ireert.get_driver(device)
+    if device == "metal" and shark_args.device_allocator == "caching":
+        print(
```
powderluv, line 551: @benvanik @antiagainst fyi there is a failure with the metal backend when using the caching allocator. We turned it on by default but that seems to fail for metal (works elsewhere).  Not a high priority for now but just something for your radar.  
@PhaneeshB please file a bug upstream to track with a repro.

PhaneeshB: https://github.com/openxla/iree/issues/14654

PR #1757, opened 2023-08-14
Remove CUDNN

gpetters94:
This removes Cuda packages and accelerate.

### DIFF
--- a/apps/language_models/langchain/langchain_requirements.txt
+++ b/apps/language_models/langchain/langchain_requirements.txt
@@ -1,3 +1,6 @@
+-f https://download.pytorch.org/whl/nightly/cpu/torch_nightly.html
+--pre
+
 # for generate (gradio server) and finetune
 datasets==2.13.0
 sentencepiece==0.1.99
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -1,3 +1,6 @@
+-f https://download.pytorch.org/whl/nightly/cpu/torch_nightly.html
+--pre
+
```
powderluv, line 3: do we need this here ? Wont torch-mlir bring in the right version ?

apps/language_models/langchain/langchain_requirements.txt
```diff
@@ -1,3 +1,6 @@
+-f https://download.pytorch.org/whl/nightly/cpu/torch_nightly.html
+--pre
+
```
powderluv, line 3: we should use the version torch-mlir is using and use that link

requirements.txt
gpetters94: Without it something is pulling in Cuda stuff, and I'm not sure what. I can set up a script to knock out lines until I find what's doing it.

vivekkhandelwal1:
@powderluv, could you please approve this? With this, the Nvidia packages are not getting installed. For Langchain, I will raise a follow-up patch. It's already disabled.

requirements.txt
vivekkhandelwal1: Many of the packages like brevitas, timm, pytorch-lightning, require a torch. If we don't add this then they won't fetch the CPU torch, resulting in Nvidia packages getting fetched. This won't result in an issue since the torch-mlir gets installed after this, and it would install the appropriate torch package. @powderluv

vivekkhandelwal1 (approved):

vivekkhandelwal1:
Sorry, I forgot to do squash.

PR #1769, opened 2023-08-16
Llama65B patch for int4 fp32

Abhishek-Varma:
Adding a WIP patch over @dan-garvey 's patch [here](https://github.com/nod-ai/SHARK/pull/1767) for `int4/fp32`.

Apart from the bug fixes, I have also changed the order of compilation (Second and then First Vicuna/Llama).

Have put it on compilation - **_if IR gets generated_** we should be able to use the `diff` of this patch on top of @dan-garvey 's patch.


Abhishek-Varma:
Would need to get rid of the hardcoded dumps of IR post compilation's success before getting the `diff` of this patch in.
CC: @dan-garvey

PR #1770, opened 2023-08-16
Add support for ROCM (Windows) in Studio + compile utils

monorimet:
TODO: Add README for setup. Enable ROCM in Windows SRT .whl builds.


apps/stable_diffusion/web/ui/stablelm_ui.py
```diff
@@ -194,6 +194,20 @@ def chat(
...
+                device = "cpu-sync"
+            elif "task" in device:
+                device = "cpu-task"
+            elif "vulkan" in device:
+                device = "vulkan"
+            elif "rocm" in device:
+                device = "rocm"
+            else:
+                print("unrecognized device")
+
```
powderluv, line 210: can we let this fall through ? There may be other vendor backends.

powderluv (approved):

PR #1780, opened 2023-08-21
Use internal dialects for globals and noinline large tensors.

stellaraccident:
This sidesteps several issues with the ml_program translation, the least of which is that it isn't preserving the noinline attribute, which we need for the moment.


stellaraccident:
(this is for reference - I think you already have something in flight)

PR #1796, opened 2023-08-25
Pipe through a debug option to iree compile utils.

monorimet:
This might seem a little heavy-handed but I've wanted to have an option piped down for things like this anyways.

### DIFF
--- a/apps/language_models/src/pipelines/SharkLLMBase.py
+++ b/apps/language_models/src/pipelines/SharkLLMBase.py
@@ -3,11 +3,16 @@ from abc import ABC, abstractmethod
 
 class SharkLLMBase(ABC):
     def __init__(
-        self, model_name, hf_model_path=None, max_num_tokens=512
+        self,
+        model_name,
+        hf_model_path=None,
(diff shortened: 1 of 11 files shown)


apps/language_models/src/pipelines/SharkLLMBase.py
```diff
@@ -3,11 +3,16 @@
 
 class SharkLLMBase(ABC):
     def __init__(
-        self, model_name, hf_model_path=None, max_num_tokens=512
+        self,
+        model_name,
+        hf_model_path=None,
+        max_num_tokens=512,
```
powderluv, line 9: Why hard code this ?

powderluv (dismissed):

apps/language_models/src/pipelines/SharkLLMBase.py
monorimet: Set to false by default for minimal implementations, but I don't think the changes to LLMBase are necessary after all.

powderluv (approved):

PR #1797, opened 2023-08-25
Fix llama2 13b crashing + add spec file for CLI execution of Llama

Abhishek-Varma:
This PR adds 2 separate commits :-
1. Fixing issues with downloading/crashing of Llama2 13B - previously I was adding another flag but later saw that there is an unused one + had to change the order of downloading of the files.
2. A spec file to run CLI exe of Llama2 - the CLI spec file for SD misses out on couple of dependencies needed for Llama, so I created a separate one to not disturb the SD path.


powderluv (approved):
I'll rename llama to LLM. But can be in follow on

PR #1817, opened 2023-09-05
Code refactoring 

tosemml:
These slight code adjustments aim to enhance Pythonic quality by incorporating python `str.join`


monorimet (changes requested):
Can't we remove `res = f"[{res[:-2]}]"` and just use `res = ", ".join(ext_list)` ? seems like we add and remove a comma at the end which doesn't need to happen anymore.

dan-garvey (approved):

dan-garvey:
Looks good to me. Test failure seems unrelated

monorimet (approved):
LGTM

PR #1822, opened 2023-09-06
Remove hf_auth_token use

Abhishek-Varma:
-- This commit removes `--hf_auth_token` uses from vicuna.py.
-- It adds llama2 models based on daryl49's HF.

### DIFF
--- a/apps/language_models/scripts/vicuna.py
+++ b/apps/language_models/scripts/vicuna.py
@@ -111,12 +111,6 @@ parser.add_argument(
     help="Specify which model to run.",
 )
 parser.add_argument(
-    "--hf_auth_token",
-    type=str,
-    default=None,
-    help="Specify your own huggingface authentication tokens for models like Llama2.",
-)
-parser.add_argument(
     "--cache_vicunas",
     default=False,
     action=argparse.BooleanOptionalAction,
@@ -460,10 +454,6 @@ class ShardedVicuna(VicunaBase):
 
     def get_tokenizer(self):
         kwargs = {}
-        if self.model_name == "llama2":
-            kwargs = {
-                "use_auth_token": "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"
-            }
         tokenizer = AutoTokenizer.from_pretrained(
             self.hf_model_path,
             use_fast=False,
@@ -1217,7 +1207,6 @@ class UnshardedVicuna(VicunaBase):
         self,
         model_name,
         hf_model_path="TheBloke/vicuna-7B-1.1-HF",
-        hf_auth_token: str = None,
         max_num_tokens=512,
         device="cpu",
         precision="int8",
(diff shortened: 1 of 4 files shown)


Abhishek-Varma:
Currently marking it as draft since 13B and 70B paths need testing.
CC: @powderluv

powderluv:
If we only download the mlir we wouldn't hit the token right?

Abhishek-Varma:
> If we only download the mlir we wouldn't hit the token right?

I did try doing that but during the run saw that we will hit that issue - because we're using [tokenizers to decode each generated token](https://github.com/nod-ai/SHARK/blob/main/apps/language_models/scripts/vicuna.py#L1669-L1677). And [this tokenizer](https://github.com/nod-ai/SHARK/blob/main/apps/language_models/scripts/vicuna.py#L1278-L1285) is being instantiated as per the HF repo we use.

Abhishek-Varma:
> > If we only download the mlir we wouldn't hit the token right?
> 
> I did try doing that but during the run saw that we will hit that issue - because we're using [tokenizers to decode each generated token](https://github.com/nod-ai/SHARK/blob/main/apps/language_models/scripts/vicuna.py#L1669-L1677). And [this tokenizer](https://github.com/nod-ai/SHARK/blob/main/apps/language_models/scripts/vicuna.py#L1278-L1285) is being instantiated as per the HF repo we use.

Even this would work since we're anyway blocking the IR generation.
It'd then essentially download the tokenizer's config files from [daryl149/llama-2-7b-hf](https://huggingface.co/daryl149/llama-2-7b-hf) and we already have the MLIR generated from [meta-llama/Llama-2-7b-chat-hf](https://huggingface.co/meta-llama/Llama-2-7b-chat-hf).

I verified it on CPU for llama2 7B.

With this PR we don't need to maintain config files for tokenizer but we're changing the base HF repo and this would impact the workflow when the IR generation is given a green signal.

But with [the other PR](https://github.com/nod-ai/SHARK/pull/1824) we only need to incur an overhead for maintaining the config files - keeping rest of the infra same.

apps/language_models/scripts/vicuna.py
```diff
@@ -1851,9 +1822,9 @@ def create_prompt(model_name, history):
 
     model_list = {
         "vicuna": "vicuna=>TheBloke/vicuna-7B-1.1-HF",
-        "llama2_7b": "llama2_7b=>meta-llama/Llama-2-7b-chat-hf",
-        "llama2_13b": "llama2_13b=>meta-llama/Llama-2-13b-chat-hf",
-        "llama2_70b": "llama2_70b=>meta-llama/Llama-2-70b-chat-hf",
+        "llama2_7b": "llama2_7b=>daryl149/llama-2-7b-chat-hf",
+        "llama2_13b": "llama2_7b=>daryl149/llama-2-13b-chat-hf",
+        "llama2_70b": "llama2_7b=>daryl149/llama-2-70b-chat-hf",
```
vivekkhandelwal1, line 1827: I think there's a typo here. All are named 7b.

vivekkhandelwal1 (changes requested):

vivekkhandelwal1 (approved):

PR #1828, opened 2023-09-11
(SD) Fix tokenizers imports in pyinstaller builds.

monorimet:
Also disables vae tuning for RDNA3.

### DIFF
--- a/apps/stable_diffusion/shark_studio_imports.py
+++ b/apps/stable_diffusion/shark_studio_imports.py
@@ -52,6 +52,7 @@ datas += collect_data_files("jsonschema")
 datas += collect_data_files("jsonschema_specifications")
 datas += collect_data_files("cpuinfo")
 datas += collect_data_files("langchain")
+datas += collect_data_files("cv2")
 datas += [
     ("src/utils/resources/prompts.json", "resources"),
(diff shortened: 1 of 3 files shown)


gpetters94 (commented):
LGTM

dan-garvey (commented):
why does it not matter that we get rid of gcs and opencv?

monorimet:
> why does it not matter that we get rid of gcs and opencv?

see https://github.com/nod-ai/SHARK/actions/runs/6142135995/job/16663459666#step:6:1042

```
2498 WARNING: collect_data_files - skipping data collection for module 'opencv_python' as it is not a package.
2737 WARNING: collect_data_files - skipping data collection for module 'google_cloud_storage' as it is not a package.
```

meaning we can remove them as they've always been skipped

PhaneeshB:
> meaning we can remove them as they've always been skipped (at least the shark_downloader doesn't seem to need gcs. I'm not sure about opencv but hasn't posed issues yet.)


Not sure why exe is skipping opencv, iirc,
OpenCV is imported for one of the stencil modes

PhaneeshB (approved):
LGTM!

PR #1829, opened 2023-09-12
[Llama2] Add fix for generating past key values

Abhishek-Varma:
-- torch.tensor on list of np.arrays is VERY SLOW.
-- This commit therefore converts the list to a np.array and then uses torch.tensor on the same.

This therefore solves the issue of indefinite hanging after the First Llama is invoked.

### DIFF
--- a/apps/language_models/scripts/vicuna.py
+++ b/apps/language_models/scripts/vicuna.py
@@ -8,6 +8,7 @@ from tqdm import tqdm
 from typing import List, Tuple
 import subprocess
 import sys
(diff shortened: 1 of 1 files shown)


vivekkhandelwal1 (approved):

Abhishek-Varma:
@Shukla-Gaurav you can cherry-pick this.
This solves the issue of Vulkan execution getting stuck indefinitely.
I was able to get the tokens to generate on both CLI and WebUI.

PR #1834, opened 2023-09-18
Add ZoeDepth

gpetters94:


apps/stable_diffusion/src/utils/stencils/zoe/__init__.py
```diff
@@ -0,0 +1,69 @@
...
+
+        if not modelpath.is_file():
+            midas = torch.hub.load(
+                "gpetters94/MiDaS",
+                "DPT_BEiT_L_384",
+                pretrained=False,
+                force_reload=False,
+            )
+            zoe = torch.hub.load(
+                "AyaanShah2204/MiDaS",
```
AyaanShah2204, line 37: @gpetters94 why are we calling the MiDaS model twice? Did you mean `AyaanShah2204/ZoeDepth`?

AyaanShah2204: Also, I can modify `AyaanShah2204/ZoeDepth` to internally load `Ayaanshah2204/MiDaS` instead of the `intel-org` version. Then we don't need to worry about `torch.load` potentially caching conflicting models.

gpetters94: > Also, I can modify `AyaanShah2204/ZoeDepth` to internally load `Ayaanshah2204/MiDaS` instead of the `intel-org` version. Then we don't need to worry about `torch.load` potentially caching conflicting models.

That'd be great, that was causing some issues. If you could also rebase your MiDaS version with mine we can switch yours back. It's the fix you were talking about in the MiDaS repo (the one Illyasviel did) that fixed the version issues with TOM SHARK.

Abhishek-Varma:
Hi @gpetters94 - I've added a similar comment to [normal stencil](https://github.com/nod-ai/SHARK/pull/1842#pullrequestreview-1644730400).

Can we look for a minimal inclusion for both these features? Since these two features would itself take ~14K new lines of code in SHARK - most of which we don't need and adds to the overhead!

Abhishek-Varma:
PTAL at [controlnet-hinter](https://github.com/takuma104/controlnet_hinter/tree/main/src/annotator).

Abhishek-Varma (commented):
Looks good now! Thanks for incorporating the changes.
Please check if this works in exe because of `einops` dependency.

apps/stable_diffusion/src/utils/stencils/zoe/__init__.py
```diff
@@ -0,0 +1,64 @@
...
+        ckpt_path.mkdir(parents=True, exist_ok=True)
+        modelpath = ckpt_path / "ZoeD_M12_N.pt"
+
+        with requests.get(remote_model_path, stream=True) as r:
+            r.raise_for_status()
+            with open(modelpath, "wb") as f:
+                for chunk in r.iter_content(chunk_size=8192):
+                    f.write(chunk)
+
+        midas = torch.hub.load(
```
AyaanShah2204, line 27: - `AyaanShah2204/ZoeDepth` is configured to trigger a `torch.hub.load` for `AyaanShah2204/MiDaS`. (This happens in the `./zoedepth/models/base_models/midas.py` line 341) 
- `AyaanShah2204/MiDaS` has the necessary changes to not require a old versions of `timm`
- Theoretically, if we were to remove this line of code, it still would work and we would be able to unpin MiDaS

I don't have SHARK set up on my local machine so I have no way to verify this.

monorimet (approved):
LGTM!

monorimet (approved):
LGTM.

PR #1835, opened 2023-09-18
Switch to keras-nightly to fix the Linux CI issue

gpetters94:


gpetters94:
This should fix the CI issue on Linux (it works on my system at least, if it doesn't work on the CI then we can pin an older keras stable version that also works.

monorimet (approved):
LGTM

PR #1841, opened 2023-09-21
[Llama2] Fix wrong Vulkan device ID + Add Vulkan compile flags

Abhishek-Varma:
-- This commit fixes the wrong Vulkan device being selected during
   runtime.
-- It also adds couple of IREE compilation flags to target specific
   Vulkan device.
-- It also changes the Vulkan device listing to be in tune with the lowering
    control flow.
    
### DIFF
--- a/apps/language_models/scripts/vicuna.py
+++ b/apps/language_models/scripts/vicuna.py
@@ -1229,6 +1229,7 @@ class UnshardedVicuna(VicunaBase):
         download_vmfb=False,
         cache_vicunas=False,
         extra_args_cmd=[],
+        device_id=None,
         debug=False,
     ) -> None:
         super().__init__(
@@ -1247,6 +1248,7 @@ class UnshardedVicuna(VicunaBase):
         print(f"[DEBUG] hf model name: {self.hf_model_path}")
         self.max_sequence_length = 256
         self.device = device
+        self.device_id = device_id
         self.precision = precision
         self.download_vmfb = download_vmfb
         self.vicuna_vmfb_path = vicuna_vmfb_path
@@ -1409,7 +1411,7 @@ class UnshardedVicuna(VicunaBase):
                 single_file=True,
             )
         self.shark_model = get_vmfb_from_path(
-            self.vicuna_vmfb_path, self.device, "tm_tensor"
+            self.vicuna_vmfb_path, self.device, "tm_tensor", self.device_id
         )
         if self.shark_model is not None:
             print(f"[DEBUG] vmfb found at {self.vicuna_vmfb_path.absolute()}")
@@ -1657,6 +1659,7 @@ class UnshardedVicuna(VicunaBase):
             mlir_module=combined_module,
             device=self.device,
             mlir_dialect="tm_tensor",
+            device_idx=self.device_id
         )
         path = shark_module.save_module(
             self.vicuna_vmfb_path.parent.absolute(),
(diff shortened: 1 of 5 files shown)


Abhishek-Varma:
We can make this concise as I mentioned in the TODO section in a follow-up patch.

I've tested with this patch and reported latest numbers on perf of llama2 vulkan.

apps/language_models/scripts/vicuna.py
```diff
@@ -1808,11 +1811,39 @@ def create_prompt(model_name, history):
...
+                break
+            id += 1
+        
+        assert device_id, f"no vulkan hardware for target-triple '{vulkan_target_triple}' exists"
+        # Step 2. Add a few flags targetting specific hardwares.
+        if "rdna" in vulkan_target_triple:
+            flags_to_add = [
+                "--iree-spirv-index-bits=64",
+                "--iree-opt-const-expr-hoisting=false",
+                "--iree-scheduling-optimize-bindings=false"
```
qedawkins, line 1843: We should only require "--iree-spirv-index-bits=64". The other two were to help narrow down the set of possible issues.

Abhishek-Varma: Oh okay. I've addressed it.

apps/stable_diffusion/web/ui/stablelm_ui.py
```diff
@@ -169,10 +171,53 @@ def chat(
...
+            )
+            _extra_args.append(
+                f"-iree-vulkan-target-triple={vulkan_target_triple}"
+            )
+            if "rdna" in vulkan_target_triple:
+                flags_to_add = [
+                    "--iree-spirv-index-bits=64",
+                ]
+                _extra_args = _extra_args + flags_to_add
+        elif device == "vulkan" and vulkan_target_triple != "":
```
qedawkins, line 195: nit: Condition on `device == "vulkan"` once and fetch the vulkan target triple if not already set. Then you should only have to specify the extra flags to add once.

qedawkins (dismissed):
The model changes look good to me (after a small refactor). Maybe wait for someone else to comment on the device selection stuff.

~~Also I would make sure the CI failures are unrelated.~~ Looks like that's from the maxf -> maximumf changes. You can probably ignore the CI there.

apps/stable_diffusion/web/ui/stablelm_ui.py
Abhishek-Varma: I've made the appropriate change - you may re-review because the `push` of the update dismissed your review. :-P

qedawkins: I feel like we should change that setting on the repo. Not sure that having to restamp over and over again is helpful :/

qedawkins (approved):

Shukla-Gaurav:
@Abhishek-Varma Do we really need an extra device_id parameter and target triple logic to match the right device_id?
Seems like this patch handles mapping to correct device when there are multiple vulkan devices(and extra flags for AMD device), correct me if I am wrong?
 
AFAIK we have handled this problem(mapping to right device) already for stable diffusion in a simpler way I guess. Can you please cross check once?  Thanks!

Abhishek-Varma:
> @Abhishek-Varma Do we really need an extra device_id parameter and target triple logic to match the right device_id? Seems like this patch handles mapping to correct device when there are multiple vulkan devices(and extra flags for AMD device), correct me if I am wrong?
> 
> AFAIK we have handled this problem(mapping to right device) already for stable diffusion in a simpler way I guess. Can you please cross check once? Thanks!

The mapping performed in StableDiffusion is not the correct way we get hold of [vulkaninfo](https://github.com/nod-ai/SHARK/blob/main/shark/iree_utils/vulkan_utils.py#L27-L38). I have indeed changed [that](https://github.com/Abhishek-Varma/SHARK/blob/c50977d41f8cb64e1cd67b82e70ecb291991e4d0/apps/stable_diffusion/src/utils/utils.py#L478-L487) in this patch. PTAL.

The mapping is completely opposite of what we get during the actual compilation stage.
Although I don't have the screenshot for Windows - it'd be fairly easy to see what I'm pointing towards because there the essence of multiple Vulkan hardwares can be seen.

For Linux :-
Incorrect/Missing info w.r.t Vulkaninfo :-
Screenshot 2023-09-21 at 6 23 39 PM

Correct (with this patch you'd see correct listing and therefore correct mapping) :-
Screenshot 2023-09-21 at 6 26 42 PM

In case of Windows it maps `rdna3` with `0` and `ampere` with `1`, whereas Vulkaninfo fetches us exactly opposite device ID map and it breaks the compilation/runtime. PTAL!

Abhishek-Varma:
@Shukla-Gaurav - I've addressed your query above (forgot to tag so adding this as a ping).

Abhishek-Varma:
Getting this in (since this patch works for multiple Vulkan hardwares) - can address any concerns in a follow-up patch

PR #1847, opened 2023-09-27
Improve mem management in opt_perf_comparison.py

monorimet:
This deletes the huggingface model objects after import_with_fx and before .vmfb compilation, to address memory issues.


dan-garvey (approved):

monorimet:
Closed -- will adress in #1855

PR #1849, opened 2023-09-28
Modify falcon pipeline for 180b support

vivekkhandelwal1:


apps/language_models/src/pipelines/falcon_pipeline.py
```diff
@@ -74,33 +83,48 @@ def __init__(
...
         self.src_model = self.get_src_model()
+        self.shark_model = self.compile()
 
     def get_tokenizer(self):
         tokenizer = AutoTokenizer.from_pretrained(
-            self.hf_model_path, trust_remote_code=True
+            self.hf_model_path,
+            trust_remote_code=True,
+            token=self.hf_auth_token,
         )
```
Abhishek-Varma, line 110: Can we get rid of this with a similar approach to what's being done for llama in `vicuna.py`?
This in turn should mean that we can perhaps get rid of `hf_auth_token`?

vivekkhandelwal1:
The CI failure is unrelated to the changes made in this PR.

apps/language_models/src/pipelines/falcon_pipeline.py
```diff
@@ -189,12 +215,18 @@ def compile_falcon(self):
...
                 module.operation.write_bytecode(bytecode_stream)
                 bytecode = bytecode_stream.getvalue()
-                del module
 
                 print(f"[DEBUG] writing mlir to file")
-                with open(f"{self.model_name}.mlir", "wb") as f_:
+                with open(f"{self.falcon_mlir_path}_elided.mlir", "w") as f_:
                     with redirect_stdout(f_):
-                        print(module.operation.get_asm())
+                        print(module.operation.get_asm(large_elements_limit=4))
```
Abhishek-Varma, line 222: For debugging purpose it's fine but I don't think we should enable dumping elided IR via any control path.

apps/language_models/src/pipelines/falcon_pipeline.py
```diff
@@ -189,12 +215,18 @@ def compile_falcon(self):
...
+                f_.close()
+
+                del module
+
+                f_ = open(self.falcon_mlir_path, "wb")
+                f_.write(bytecode)
+                print("Saved mlir")
```
Abhishek-Varma, line 229: Better print the name of the IR after saving it.

apps/language_models/src/pipelines/falcon_pipeline.py
```diff
@@ -204,11 +236,10 @@ def compile_falcon(self):
             self.falcon_vmfb_path.parent.absolute(),
             self.falcon_vmfb_path.stem,
             extra_args=[
-                "--iree-hal-dump-executable-sources-to=ies",
+                f"--iree-hal-dump-executable-sources-to=hal_sources_{self.model_name}",
```
Abhishek-Varma, line 239: I've seen this in a few more scripts like [MiniGPT4](https://github.com/nod-ai/SHARK/blob/4fec03a6ccca77d53c2ee3d6da3453c19b5056c8/apps/language_models/src/pipelines/minigpt4_pipeline.py#L301) and I feel we should get rid of this flag OR perhaps add it based on a `--debug` flag.

Abhishek-Varma (commented):
Hi @vivekkhandelwal1 - I've reviewed and added a few comments.

apps/language_models/src/pipelines/falcon_pipeline.py
vivekkhandelwal1: I don't think the same can be done for Falcon now. Since the Falcon pipeline needs the original model during the inference as well. Also, the token is required just for the falcon-180b which won't be used once I add another patch for the 180b-gptq variant because that doesn't require any auth token.

Abhishek-Varma (approved):
LGTM.

PR #1855, opened 2023-09-29
Add --additional_runtime_args option and use in OPT example.

monorimet:
Fix the func name. (#1838)

fix non-plugin path

fixup OPT example

Use compile_file


tank/examples/opt/opt_perf_comparison.py
```diff
@@ -96,39 +119,53 @@ def create_vmfb_module(
...
-            f.write(model_mlir)
-        print(f"Saved mlir at {mlir_path}")
-
-    shark_module = SharkInference(
-        model_mlir,
-        device=device,
-        mlir_dialect="tm_tensor",
-        is_benchmark=False,
-    )
+    # with open(mlir_path, "r") as f:
```
dan-garvey, line 129: dead code?

tank/examples/opt/opt_perf_comparison.py
```diff
@@ -96,39 +119,53 @@ def create_vmfb_module(
...
+    #    rt_flags=[],
+    # )
 
     vmfb_name = (
         f"{opt_fs_name}_causallm_{max_seq_len}_torch_{DEVICE}_tiled_ukernels"
     )
-    shark_module.save_module(module_name=vmfb_name)
+    # shark_module.save_module(module_name=vmfb_name)
     vmfb_path = vmfb_name + ".vmfb"
+    ireec.compile_file(
```
dan-garvey, line 146: we dont have an abstraction for this?

dan-garvey (changes requested):

tank/examples/opt/opt_perf_comparison.py
monorimet: This was proof of concept for the new path through shark inference (compile_str --> compile_file).
Updated it, thanks!

tank/examples/opt/opt_perf_comparison.py
monorimet: See next comment :)

dan-garvey (approved):

PR #1856, opened 2023-09-30
Add flag for specifying device-local caching allocator heap key.

monorimet:
This is really just a temporary solution to piping these kinds of IREE runtime options from SHARK Studio.
We should probably just allow the specification of `iree_runtime_args` to be piped down in a device-agnostic fashion instead of naming them here in stable_args or in shark_args (worse), though it does help to have a curated list of knobs to tweak for users.

Currently only piped through for Vulkan devices.

I have a better implementation of supplying runtime args to the core SHARK python runtime utils in progress here: https://github.com/nod-ai/SHARK/pull/1855 but it still needs some work.


qedawkins (approved):
LGTM!

PR #1862, opened 2023-10-03
[vulkan] Print note about module load times. NFC.

kuhar:
Print a note ahead of a potentially long inactivity to set the right expectations.

Separately, it seems possible to make this loading significantly faster.

### DIFF
--- a/shark/iree_utils/compile_utils.py
+++ b/shark/iree_utils/compile_utils.py
@@ -406,7 +406,9 @@ def load_vmfb_using_mmap(
             dl.log(f"ireert.SystemContext created")
             if "vulkan" in device:
                 # Vulkan pipeline creation consumes significant amount of time.
-                print("\tCompiling Vulkan shaders. This may take a few minutes.")
+                print(
+                    "\tCompiling Vulkan shaders. This may take a few minutes."
+                )
             ctx.add_vm_module(mmaped_vmfb)
             dl.log(f"module initialized")
             mmaped_vmfb = getattr(ctx.modules, mmaped_vmfb.name)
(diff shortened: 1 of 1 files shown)


shark/iree_utils/compile_utils.py
```diff
@@ -358,6 +358,10 @@ def load_vmfb_using_mmap(
     flatbuffer_blob_or_path, device: str, device_idx: int = None
 ):
     print(f"Loading module {flatbuffer_blob_or_path}...")
+    if "vulkan" in device:
+        # Vulkan pipeline creation consumes significant amount of time.
+        print("(This step may take a few minutes)")
```
qedawkins, line 363: nit: You could specify `compiling shaders` or something similar.

qedawkins (commented):
I think this is reasonable, but this reporting won't appear in the web interface, and I imagine that's where it is most critical.

qedawkins:
(also you can ignore all of the CI except the linter which passed)

kuhar:
> I think this is reasonable, but this reporting won't appear in the web interface, and I imagine that's where it is most critical.

Yes, we need both. This is just the tiniest first step towards better loading experience.

shark/iree_utils/compile_utils.py
kuhar: Thanks. I added this and moved the message closer to where the compilations happen.

qedawkins (approved):

kuhar:
Print a note ahead of a potentially long inactivity to set the right expectations.

Separately, we should add progress to the UI and make this loading faster.

PR #1899, opened 2023-10-17
Fix MacOS web execution flow

PhaneeshB:
- drivers with single device will not be shown with `://0` indexing
- plumb metal device through chatbot execution (currently the default metal device is selected)


monorimet (approved):
LGTM.

PR #1900, opened 2023-10-18
update dependency on rocm/hip info command

PhaneeshB:
- temp fix to pass rocm device architecture using `--iree_rocm_target_chip=<arch>`

### DIFF
--- a/shark/iree_utils/_common.py
+++ b/shark/iree_utils/_common.py
@@ -120,14 +120,8 @@ def check_device_drivers(device):
     elif device == "cpu":
         return False
     elif device == "rocm":
-        try:
-            if sys.platform == "win32":
-                subprocess.check_output("hipinfo")
-            else:
-                subprocess.check_output("rocminfo")
-        except Exception:
-            return True
-
+        print(f"Device ROCM - Assuming hip / rocm libs already installed.")
+        return False
     # Unknown device. We assume drivers are installed.
     return False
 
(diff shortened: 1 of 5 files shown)


stellaraccident:
Works for me.

For a more future looking approach, we should:

* add iree runtime support that lets us get the chip value for an enumerated device
* enable multi targeting and compile for a set of default supported chips

monorimet (commented):
How does the bitcode path get exposed to iree-compile if we don't get it here? Is it no longer necessary to use `--iree-rocm-bc-dir=<>`?

shark/iree_utils/_common.py
```diff
@@ -120,14 +120,8 @@ def check_device_drivers(device):
...
     elif device == "rocm":
-        try:
-            if sys.platform == "win32":
-                subprocess.check_output("hipinfo")
-            else:
-                subprocess.check_output("rocminfo")
-        except Exception:
-            return True
-
+        print(f"Device ROCM - Assuming hip / rocm libs already installed.")
```
powderluv, line 123: we don't need rocm libs.

PhaneeshB:
> How does the bitcode path get exposed to iree-compile if we don't get it here? Is it no longer necessary to use `--iree-rocm-bc-dir=<>`?

yes, it is no longer required since IREE no comes with the necessary bc files internally.

shark/iree_utils/_common.py
PhaneeshB: Updated.

Shukla-Gaurav (approved):
lgtm

PR #1907, opened 2023-10-21
SD - Fix civitai download on Windows +other civitai download improvements

one-lithe-rune:
### Motivation

While testing https://github.com/nod-ai/SHARK/pull/1906 I found that entering Civitai api download urls into the webui wasn't working for me. This was due to the code trying to launch a subprocess to `wget` which is not guaranteed to be (and probably isn't) installed on windows, so everything falls over there.

Once I had something working in Python to do the download instead, I discovered a number of other problems with the civitai download, mainly to do with naming being taken from the civitai id number rather than the correct filename of the model file.

This PR should address most of this.

### Changes

* Move civitai model download into Python code rather than make a subprocess call to wget, as wget usually won't exist on Windows.
* Save model files downloaded from civitai under the correct filename reported by civitai rather than their civitai index number. i.e. 'ghostmix_v20Bakedvae.safetensors' instead of '36520.safetensors'
* If it has been set, use args.chkpt_dir as the  destination folder when downloading from civitai. Only hardcode 'models' as the fallback.
* Set args.chkpt_loc after civitai download or name resolution so webui adds the correct model name to any metadata it writes for generated images.
* Use a stricter check for determining whether custom_weights are a civitai api download url.
* Update .gitignore to only exclude the root /models folder rather than all models folders (allows files in /apps/stable_diffusion/src/models/ to be committed without git whining)

### Problems/Concerns

* The webui should ideally update itself once a civitai download has occurred to show and list the correct .safetensors file replacing the civitai download url in its interface. This though is starting to get into 'precompile' rather than 'ondemand' area and I didn't want get into that level of pipeline/webui overhaul.
* Given the flakiness of civitai in general, more explicit error handling for when it has broken would be sensible. I haven't been sensible.
* There doesn't seem to be any alternative to poking the civitai download url to get the filename from the headers to check whether we already have the .safetensors file downloaded. This doesn't feel ideal.
* I'm not sure my .gitignore change is correct. It might be too general.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -182,7 +182,7 @@ generated_imgs/
 
 # Custom model related artefacts
 variants.json
-models/
+/models/
 
 # models folder
 apps/stable_diffusion/web/models/
--- a/apps/stable_diffusion/src/models/model_wrappers.py
+++ b/apps/stable_diffusion/src/models/model_wrappers.py
@@ -8,6 +8,7 @@ import traceback
 import subprocess
 import sys
 import os
+import requests
 from apps.stable_diffusion.src.utils import (
     compile_through_fx,
     get_opt_flags,
@@ -16,6 +17,7 @@ from apps.stable_diffusion.src.utils import (
     preprocessCKPT,
     convert_original_vae,
     get_path_to_diffusers_checkpoint,
+    get_civitai_checkpoint,
     fetch_and_update_base_model_id,
     get_path_stem,
     get_extended_name,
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -182,7 +182,7 @@ generated_imgs/
 
 # Custom model related artefacts
 variants.json
-models/
+/models/
```
powderluv, line 185: did you need this change to ignore changes in models/ ? I thought the existing one should have worked.

one-lithe-rune: It was ignoring the `models` folder fine, but it also seemed to be ignoring anything with `models` as a folder in the path, and git would keep complaining about it when I tried to add my changes for 'apps/stable_diffusion/src/**models**/model_wrappers.py'

I might just be my git installation or my editor being weird.

monorimet: I also noticed this with language models, didn't think much of it but this is probably the reason why that was happening. Thanks

monorimet (approved):
Looks good to me. Thank you!!!

PR #1912, opened 2023-10-22
SD: Fix unet untuned opt_flags

one-lithe-rune:
### Motivation

I didn't make the correct change to the untuned opt_flags for unet in #1910 😬

### Changes

* correct my sloppy copy/paste for the untuned unet default compilation flags that introduced an extra 'detach' into what should have been 'iree-global-opt-convert-1x1-filter-conv2d-to-matmul'

* ROCM is actually slower with these flags than having them broken. ~1.75 it/s broken, ~1.5 it/s with this fix for a SD 1.5 based model.

### DIFF
--- a/apps/stable_diffusion/src/utils/resources/opt_flags.json
+++ b/apps/stable_diffusion/src/utils/resources/opt_flags.json
@@ -11,12 +11,12 @@
     "untuned": {
       "fp16": {
         "default_compilation_flags": [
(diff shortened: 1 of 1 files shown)


njsharpe:
I manually made this changes in my local install and there is still a slight decrease in performance than what used to be available to us by as low as 2it/s to as high as 5it/s. Where is this gap coming from?

monorimet (approved):
OK. This should be the right move -- I'll be triaging perf on rocm/vulkan and can remove this flag conditionally if needed.

PR #1916, opened 2023-10-24
Add opt_causallm_samples.py.

godot73:
- Add opt_causallm_samples.py that run opt_causallm.generate_tokens() with the sample prompts.
- Separate out the sample prompts from opt_perf_comparison.py to opt_util.py for reuse.


monorimet (approved):
LGTM.

PR #1921, opened 2023-10-27
Fix wrong parameter index passed to 'compile_module_to_flatbuffer'

no1wudi:
compile_str is always False in compile_module_to_flatbuffer since there is a parameter 'model_name' before 'debug'.

This issue is relative to https://github.com/nod-ai/SHARK/pull/1863.

Then we can use mlir model buffer in RAM to run inference.

### DIFF
--- a/shark/iree_utils/compile_utils.py
+++ b/shark/iree_utils/compile_utils.py
@@ -461,13 +461,13 @@ def get_iree_compiled_module(
 ):
     """Given a module returns the compiled .vmfb and configs"""
     flatbuffer_blob = compile_module_to_flatbuffer(
-        module,
-        device,
-        frontend,
-        model_config_path,
-        extra_args,
-        debug,
-        compile_str,
+        module=module,
+        device=device,
+        frontend=frontend,
+        model_config_path=model_config_path,
+        extra_args=extra_args,
+        debug=debug,
+        compile_str=compile_str,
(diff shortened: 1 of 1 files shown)


monorimet (approved):
Wow thank you so much for tracking this down. LGTM

shark/iree_utils/compile_utils.py
```diff
@@ -535,13 +535,13 @@ def export_iree_module_to_vmfb(
...
-        module,
-        device,
-        mlir_dialect,
-        model_config_path,
-        extra_args,
-        debug,
-        compile_str,
+        model=module,
+        device=device,
+        frontend=frontend,
```
monorimet, line 540: You can change this to `frontend=mlir_dialect` -- it's not ideal but they are used pretty interchangeably at this point and don't matter much at the moment since we are mostly using PyTorch.

monorimet (changes requested):

shark/iree_utils/compile_utils.py
monorimet: (See CPU CI - https://github.com/nod-ai/SHARK/actions/runs/6666541497/job/18118262556?pr=1921#step:8:1614 )

monorimet (approved):
LGTM. Thanks.

PR #1922, opened 2023-10-27
Add smoothquant OPT to examples.

monorimet:


dan-garvey (approved):
Cool. Any reason not to add a test for at least one of these model variants?

monorimet:
> Cool. Any reason not to add a test for at least one of these model variants?

Need to rework the turbine integration for models with multiple outputs (including OPT) so these will be included in that patch. Good idea!

PR #1924, opened 2023-10-28
SD/API: Add missing A1111 APIs to Shark to support koboldcpp image generation

one-lithe-rune:
### Motivation

[koboldcpp-1.47.2](https://github.com/LostRuins/koboldcpp/releases/tag/v1.47.2) added support for doing generation of its embedded images via the A1111 api. Shark needed some more stuff implemented to make this work.

Then it turned out I wanted less hard-coding of the arguments not passed by the API so I could get the images I wanted, and then that meant updating the other apis to match. Plus fixes and other stuff.

### Changes (so far)

* Refactors SD api functions into their own module.
* Adds the following apis implemented by A1111 (first two needed by koboldcpp):
   - adds /sdapi/v1/sd-models (lists available models)
   - adds /sdapi/v1/options (just a little more than the bare minimum needed)
   - (0d1c2fb7a5ce720e6c16263c7a417bd42a91a96e) adds /sdapi/v1/samplers (lists schedulers)
   - (0d1c2fb7a5ce720e6c16263c7a417bd42a91a96e) adds /sdapi/v1/cmd-flags (lists command line flag values)
* Adds optional CORS support, use the '--api_accept_origin' command line argument to activate and configure (required by koboldcpp).
* (0d1c2fb7a5ce720e6c16263c7a417bd42a91a96e) Adds pydantic typing for request validation and much improved API endpoint docs. 
* (0d1c2fb7a5ce720e6c16263c7a417bd42a91a96e) Change how the api is mounted to move stuff out of `index.py`. However this means the docs for /sdapi have moved. `swagger` is now at /sdapi/docs with `redoc` at /sdapi/redoc. LLM doc locations are unchanged.
* Extends existing API endpoints to include optional sampler/scheduler selection.
* (0d1c2fb7a5ce720e6c16263c7a417bd42a91a96e) Support some A1111 sampler names as aliases to the `diffusers` names we use.
* (0d1c2fb7a5ce720e6c16263c7a417bd42a91a96e) Extends existing API endpoints to be able return multiple images using the `n_iter` request property (A1111 equivalent to `--batch_count`)
* (0d1c2fb7a5ce720e6c16263c7a417bd42a91a96e) Extends txt2img API endpoints with some hires fix properties that appear in the A1111 version.
* Extends /sdapi/v1/textimg to recognise the method used by koboldcpp to select the model.
* Where possible take values not provided to the API in the request from the existing relevant command line parameters rather than hardcoding them.
* ~~return a 400 response when a request doesn't have required properties.~~ (superseded by pydantic typing)
* changed default schedulers and models for some apis to ones that actually seem to work.
* Update api_test.py to include the new API endpoints.
* Update api_test.py to include a '--verbose' command line option.
* (0d1c2fb7a5ce720e6c16263c7a417bd42a91a96e) Update api_test.py to always display response contents if the status_code was something other than '200 OK'
* (Added 7cb76a4cefd80805ba73f029df8213bcba0e4320) Include --use_lora and --device in the command lines args that the API will use if they are set.
* (0d1c2fb7a5ce720e6c16263c7a417bd42a91a96e) Changed the command line args `--bottom` and `--top` to now be aliases for `--up` and `--down` as everywhere else is using those, and it made the api typing/defaulting easier. (currently resisting the urge to add `--charm` and `--strange`)
* (5331e7e743fe6e5c1ecd9c9ed20780253a418680) Added `shark_sd_koboldcpp.md` to the `/docs` folder and linked it and the already existing Blender one into a new section of the main README

### In Action

Taking back the subsea half-kilometer - by koboldcpp


powderluv:
This is awesome!!

monorimet:
Wow!!! this is really helpful! Let us know if you have any questions - I've tagged @Shukla-Gaurav and @gpetters94 as well

one-lithe-rune:
That's everything on my list checked off, so:

### Possible Problems/Concerns/Notes

* A1111 doesn't have inpaint, outpaint, or upscaler endpoints as part of its api so those are obviously not in sync with it.  It seems to stuff a lot of inpainting things into its img2img endpoint. I didn't attempt to do the same, so img2img also still differs quite strongly.
* I've added hires_fix properties into the txt2img endpoint because A1111 has them, but I couldn't work out all the correspondences, so its a bit of mish mash between properties
* Sampler defaults shown in the swagger and redoc aren't correct, even though the actual defaults used seem to be right.
* Blender API document could do with some updates with the newer .exe naming, but I felt that was out of scope.
* I don't *think* I've broken anything by the outpaint direction argument changes, but I'm not 100% sure.
* All the /sdapi endpoints are in one file. There is an argument that it should be one endpoint per file. But I didn't really like them being split across many files previously, and I confess I didn't feel I understood the FastAPI documenation on how to do that enough to attempt it, when defining the endpoints using decorators against functions whilst mounting the whole API as a FastAPI app from index.py rather than pulling in the functions one by one.
* The API still relies on importing the _inf functions from the ui module.
* I haven't tested against anything other than Koboldcpp and the `/rest-api-tests/api-test.py` script.

one-lithe-rune:
Rebased.

monorimet (approved):
This is great. It looks good to me -- let me test this out for blender and koboldcpp and then we can merge if you're ready @one-lithe-rune

one-lithe-rune:
> This is great. It looks good to me -- let me test this out for blender and koboldcpp and then we can merge if you're ready @one-lithe-rune

I think its good to go.

PR #1934, opened 2023-11-02
fix vicuna cli vulkan path

PhaneeshB:
- add fixes for device selection on vulkan
- also this removes the forced ordering of devices based on device_path and now relies on device ordering from iree to be consistent across reboots (user should ensure selecting the correct device index for the current session)


monorimet (approved):
LGTM

PR #1941, opened 2023-11-06
[SDXL] Add SDXL pipeline to SHARK

Abhishek-Varma:
-- This commit adds SDXL pipeline to SHARK.

### DIFF
--- a/apps/stable_diffusion/src/models/model_wrappers.py
+++ b/apps/stable_diffusion/src/models/model_wrappers.py
@@ -1,5 +1,5 @@
 from diffusers import AutoencoderKL, UNet2DConditionModel, ControlNetModel
-from transformers import CLIPTextModel
+from transformers import CLIPTextModel, CLIPTextModelWithProjection
 from collections import defaultdict
 from pathlib import Path
 import torch
@@ -24,6 +24,8 @@ from apps.stable_diffusion.src.utils import (
     get_stencil_model_id,
     update_lora_weight,
 )
+from shark.shark_downloader import download_public_file
+from shark.shark_inference import SharkInference
 
 
 # These shapes are parameter dependent.
@@ -55,6 +57,10 @@ def replace_shape_str(shape, max_len, width, height, batch_size):
                     new_shape.append(math.ceil(height / div_val))
                 elif "width" in shape[i]:
                     new_shape.append(math.ceil(width / div_val))
+            elif "+" in shape[i]:
+                # Currently this case only hits for SDXL. So, in case any other
+                # case requires this operator, change this.
+                new_shape.append(height + width)
         else:
             new_shape.append(shape[i])
     return new_shape
@@ -65,6 +71,70 @@ def check_compilation(model, model_name):
         raise Exception(
             f"Could not compile {model_name}. Please create an issue with the detailed log at https://github.com/nod-ai/SHARK/issues"
         )
+
+
+def shark_compile_after_ir(
+    module_name,
+    device,
+    vmfb_path,
+    precision,
+    ir_path=None,
+):
+    if ir_path:
+        print(f"[DEBUG] mlir found at {ir_path.absolute()}")
+
+    module = SharkInference(
+        mlir_module=ir_path,
+        device=device,
+        mlir_dialect="tm_tensor",
+    )
+    print(f"Will get extra flag for {module_name} and precision = {precision}")
+    path = module.save_module(
+        vmfb_path.parent.absolute(),
+        vmfb_path.stem,
+        extra_args=get_opt_flags(module_name, precision=precision),
+    )
+    print(f"Saved {module_name} vmfb at {path}")
+    module.load_module(path)
+    return module
+
+
+def process_vmfb_ir_sdxl(extended_model_name, model_name, device, precision):
+    name_split = extended_model_name.split("_")
+    if "vae" in model_name:
+        name_split[5] = "fp32"
+    extended_model_name_for_vmfb = "_".join(name_split)
+    extended_model_name_for_mlir = "_".join(name_split[:-1])
+    vmfb_path = Path(extended_model_name_for_vmfb + ".vmfb")
+    if "vulkan" in device:
+        _device = args.iree_vulkan_target_triple
+        _device = _device.replace("-", "_")
+        vmfb_path = Path(extended_model_name_for_vmfb + f"_{_device}.vmfb")
+    if vmfb_path.exists():
+        shark_module = SharkInference(
+            None,
+            device=device,
+            mlir_dialect="tm_tensor",
+        )
+        print(f"loading existing vmfb from: {vmfb_path}")
+        shark_module.load_module(vmfb_path, extra_args=[])
+        return shark_module, None
+    mlir_path = Path(extended_model_name_for_mlir + ".mlir")
+    if not mlir_path.exists():
+        print(f"Looking into gs://shark_tank/SDXL/mlir/{mlir_path.name}")
+        download_public_file(
+            f"gs://shark_tank/SDXL/mlir/{mlir_path.name}",
+            mlir_path.absolute(),
+            single_file=True,
+        )
+    if mlir_path.exists():
+        return (
+            shark_compile_after_ir(
+                model_name, device, vmfb_path, precision, mlir_path
+            ),
+            None,
+        )
+    return None, None
 
 
 class SharkifyStableDiffusionModel:
@@ -86,6 +156,7 @@ class SharkifyStableDiffusionModel:
         generate_vmfb: bool = True,
         is_inpaint: bool = False,
         is_upscaler: bool = False,
+        is_sdxl: bool = False,
         use_stencil: str = None,
         use_lora: str = "",
         use_quantize: str = None,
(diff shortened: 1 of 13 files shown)


monorimet:
Cool, can't wait to try this. Let me run it quick and review

apps/stable_diffusion/scripts/txt2img.py
```diff
@@ -16,31 +16,62 @@ def main():
...
+            custom_vae=args.custom_vae,
+            low_cpu_mem_usage=args.low_cpu_mem_usage,
+            debug=args.import_debug if args.import_mlir else False,
+            use_lora=args.use_lora,
+            use_quantize=args.use_quantize,
+            ondemand=args.ondemand,
+        )
+    else:
+        assert (
+            args.height <= 768 and args.width <= 768
```
monorimet, line 54: I have an issue with splitting these up here based on the specified size.

The first thing people will do with an SDXL model is probably going to be: download SDXL model (or SDXL custom model) -> click on it in the UI dropdown -> enter prompt -> generate image

in this case they will get an error as it's set to 512x512 by default and so it just tries to use the `Text2ImagePipeline` instead of the `Text2ImageSDXLPipeline`.

Can we use the huggingface/civitAI model metadata to determine whether it is an SDXL-based model and then programmatically change the available sizes if yes? This technically works if you're aware of all the boxes to tick and dials to set, but if it breaks the first time you try it, that's bad for us.

I'm open to other points of view on this, but it could be as easy as getting the .ckpt filename and if it has XL we set some `is_sdxl` to True (False by default) and it takes the right path.


A considerable alternative to this is to move XL features to a separate `XL (Beta)` tab and then we won't have to worry about the funky stuff in the SD1.5/2.1 workflow that I ran into with this patch.

apps/stable_diffusion/src/pipelines/pipeline_shark_stable_diffusion_utils.py
```diff
@@ -306,6 +507,99 @@ def produce_img_latents(
...
+            step_start_time = time.time()
+            timestep = torch.tensor([t]).to(dtype).detach().numpy()
+            # expand the latents if we are doing classifier free guidance
+            latent_model_input = torch.cat([latents] * 2)
+
+            latent_model_input = self.scheduler.scale_model_input(
+                latent_model_input, t
+            ).to(dtype)
+
+            # input_to_unet = {
```
monorimet, line 536: Dead code? if this is debug tooling we want to keep let's flag guard it

apps/stable_diffusion/web/ui/txt2img_ui.py
```diff
@@ -156,26 +157,56 @@ def txt2img_inf(
...
+                low_cpu_mem_usage=args.low_cpu_mem_usage,
+                debug=args.import_debug if args.import_mlir else False,
+                use_lora=args.use_lora,
+                use_quantize=args.use_quantize,
+                ondemand=args.ondemand,
+            )
+        else:
+            assert (
+                height <= 768 and width <= 768
+            ), "height/width not in supported range"
```
monorimet, line 189: (Refer to above comments on handling XL - ness of a given user config)

monorimet (changes requested):
Tried this out and had a few notes. PTAL

Also, we will eventually need this on top of @aviator19941 's SD implementation in Turbine. The implementation may be slightly cleaner that way. Either way I've suggested a few changes that can make this a little more user-friendly.

Try it out yourself with a custom model, switch between XL models and SD1.5 models, stress test it a bit... and  put cushions on the rough edges and this should be ready to go.

Happy to chat and set up meetings about my comments and/or the impending Turbine integration

(FYI @aviator19941 )

Abhishek-Varma:
Hi @monorimet 

As discussed yesterday, I have addressed the following in the latest push to this revision :-
1. Added SDXL to a separate script in order to not fiddle with current normal SD.
2. Added another tab for SDXL with the set of configuration supported by SHARK SDXL.
3. Got rid of few nit changes.

It is now ready to be re-reviewed.

For CLI : `python apps/stable_diffusion/scripts/txt2img_sdxl.py --prompt="Batman playing electric guitar" --device=rocm --hf_model_id="stabilityai/stable-diffusion-xl-base-1.0"`

monorimet:
Awesome. Thank you @Abhishek-Varma. I'll give it a try later today.

monorimet:
Cool, seems to work. What needs done to have support for custom SDXL models/loras? We can merge once lint is fixed

monorimet (dismissed):

apps/stable_diffusion/scripts/txt2img.py
```diff
@@ -85,4 +85,4 @@ def main():
 
 
 if __name__ == "__main__":
-    main()
+    main()
```
monorimet, line 88: Unnecessary change maybe

apps/stable_diffusion/web/ui/txt2img_ui.py
```diff
@@ -675,4 +675,4 @@ def txt2img_inf(
                 lora_hf_id,
                 custom_vae,
             ],
-        )
+        )
```
monorimet, line 678: nit - indent change? might be fixed by black formatting

monorimet (changes requested):

apps/stable_diffusion/src/models/model_wrappers.py
```diff
@@ -67,6 +73,70 @@ def check_compilation(model, model_name):
...
+    print(f"Saved {module_name} vmfb at {path}")
+    module.load_module(path)
+    return module
+
+
+def process_vmfb_ir_sdxl(extended_model_name, model_name, device, precision):
+    name_split = extended_model_name.split("_")
+    if "vae" in model_name:
+        name_split[5] = "fp32"
+    extended_model_name_for_vmfb = "_".join(name_split)
```
monorimet, line 106: This currently appends the entire device URI to the end of the filename, resulting in filenames like "unet_1_77_1024_1024_fp16_stable-diffusion-xl-base-1_vulkan_00000000_0c00_0000_0000_000000000000.vmfb"

Can we make this conform to the naming conventions for non-SDXL models (without the device URI)? e.g.
"unet_1_64_512_512_fp16_stable-diffusion-2-1-base_vulkan.vmfb"

Abhishek-Varma: The naming of the vmfbs can be done like that but it will be wrong (and the non-SDXL models should also be updated to have the same) - a vmfb compiled for vulkan AMD device should be differentiated from vulkan NVIDIA device. Something similar is being done for Llama as well : [vicuna.py from Top-Of-Main branch](https://github.com/nod-ai/SHARK/blob/main/apps/language_models/scripts/vicuna.py#L1332-L1343), so instead of the entire URI I'm now using the `args.iree-vulkan-target-triple` flag to add the distinction.

So, now you'll see the filenames like : `clip_1_77_1024_1024_fp16_stable-diffusion-xl-base-1_vulkan_rdna2_unknown_linux.vmfb`

Abhishek-Varma:
> Cool, seems to work. What needs done to have support for custom SDXL models/loras? We can merge once lint is fixed

Hi @monorimet - I've addressed your review comments.

Regarding the custom SDXL/Loras - someone implementing that can follow the same approach used for non-SDXL models with the assumption that the HF repo contains all the sub-models and the base model being used is [stabilityai/stable-diffusion-xl-base-1.0](https://huggingface.co/stabilityai/stable-diffusion-xl-base-1.0) (since that's what we use in `base_model.json`).

apps/stable_diffusion/src/models/model_wrappers.py
```diff
@@ -86,15 +156,22 @@ def __init__(
...
-        self.height = height // 8
-        self.width = width // 8
+        self.is_sdxl = is_sdxl
+        self.height = height
+        self.width = width
+        if is_sdxl:
+            # We need to scale down the height/width by vae_scale_factor, which
+            # happens to be 8 in this case.
+            self.height = height // 8
+            self.width = width // 8
```
monorimet, line 174: This seems unintentional. The original SD needs `height // 8` and `width //  8`  and this changes the default while assigning the old behavior to SDXL.

If I try SD1.4 from scratch on this commit, I get a 4TB CPU allocation error. If I add `not` to `if is_sdxl` on line 170, the problem seems fixed for SD1.5/2.1 but now the VAE issue starts to make more sense...

I'll update after a bit more debugging.

monorimet: Oh, yeah I think it might be a good idea to just revert this one change entirely.... unless i'm mistaken

Abhishek-Varma: Perhaps following would help clear why `// 8` (`vae_scale_factor`) needs to be done for SDXL :-

1. User only mentions the final image size to be generated, be it SD1.X/2.1/XL.
2. In case of SDXL we currently only deal with 1024x1024 image generation.
3. But the inputs to UNET/VAE for a 1024x1024 image needs to be 128x128. And this is nothing but `1024 // 8`.
4. This is something I observed in the normal Python execution by printing the input sizes expected by individual models right before their "forward" method is invoked.
5. Since the inputs to each model is worked out currently via `base_model.json` and for SD1.X/2.1 it substitutes the final height/width - but for SDXL it needs to be the scaled down version (as seen in point 3 above).
6. So, I added that and guarded it with `if is_sdxl` flag.

monorimet: Before this commit, the SD1.X / 2.1 models already do `height // 8` and `width // 8` here. This commit changes the behavior for those models at this point to `height` and `width` (introduces the CPUAllocator issue) and then defines a special case for XL with the old default behavior. Unless we are reworking the pipeline for SD1.X/2.1 let's keep the non-SDXL path as it was beforehand, to reduce breakages

monorimet (approved):

PR #1943, opened 2023-11-07
Fix HipConfigError

vivekkhandelwal1:


nirvedhmeshram:
I think we should change the user script to allow passing the arch flag rather then changing the default architecture, afaik `gfx1100 ` is more common than `gfx90a` ? Just concerned we will run into same issue when trying a different architecture..

one-lithe-rune:
I assume the long term solution would be to get this from the `iree.runtime`, as you can do this on the commmand line, and the value we need is then listed under `gpu-arch-name`:


```
(shark.venv) PS C:\develop\SHARK>  iree-run-module --dump_devices=rocm
# ============================================================================
# Enumerated devices for driver 'rocm'
# ============================================================================

# ===----------------------------------------------------------------------===
# --device=rocm://GPU-00000000-0c00-0000-0000-000000000000
#   AMD Radeon RX 7900 XTX
# ===----------------------------------------------------------------------===

- gpu-compute-capability: 11.0
- gpu-arch-name: gfx1100

- launch-max-block-dims: (1024, 1024, 1024)

- block-max-thread-count: 1024
- block-max-32-bit-register-count: 0
- block-max-shared-memory: 64 KB

- memory-is-integrated-memory: 0
- memory-supports-managed-memory: 0
- memory-total-const-memory-size: 2047 MB
- memory-total-global-memory-size: 24560 MB
- memory-l2-cache-size: 4194304 bytes

- gpu-compute-unit-count: 48
- gpu-compute-max-clock-rate: 2371 mHz
- gpu-memory-max-clock-rate: 1250 mHz
- gpu-warp-size: 32
```

Don't know how to get that out of the SRT/IREE python bindings, or if its even exposed currently I'm guessing no, or it would already have been added when the querying of `hipinfo` was removed.

vivekkhandelwal1:
Okay, closing this PR based on the above comment and discussion.

PR #1946, opened 2023-11-08
llama2 - FP16 - IR generation 

PhaneeshB:
- fix combine mlir for llama2 for fp16 IR generation by keeping deduplicating constant variable names using appropriate suffix for first llama and second llama
- needs to be tested for IR generation with int4 to see it doesn't break the generation flow for quantised IRs


PhaneeshB:
Converted to draft till sanity check is done for int4 IR generation.

PR #1950, opened 2023-11-08
fix rocm arch selection

PhaneeshB:
- add support for getting rocm arch using `iree-run-module --dump_devices=rocm`
- assumption - order of devices in `iree-run-module --dump_devices=rocm` and `iree.runtime.get_driver("rocm").query_available_devices()` is the same on the system.
- the filtering of the dump info is heavily dependent on its format.

### DIFF
--- a/shark/iree_utils/_common.py
+++ b/shark/iree_utils/_common.py
@@ -19,9 +19,12 @@ import sys
 import subprocess
 
 
-def run_cmd(cmd, debug=False):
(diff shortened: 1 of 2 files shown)


vivekkhandelwal1 (approved):

JRafael000:
@PhaneeshB Hello, could you be a little clearer with the instructions? I'm having this same problem and I can't figure out how to fix it. Thank you so much.

Issue #1937, opened 2023-11-04
[pywebview] MSHTML is deprecated. See https://pywebview.flowrl.com/guide/renderer.html#web-engine on details how to use Edge Chromium


sz1kormar:
When I downloaded it the GUI won't open the right way and instead gets a white screen with no content inside of it.
As a result, I had to manually open it from my browser to see the GUI client.
![issue](https://github.com/nod-ai/SHARK/assets/96243017/40515d53-7d92-48be-8f7f-9cb728c33dd0)

henk717:
Going to give you a very quirky solution for this from my sysadmin experience.
Press Ctrl + Windows Key + Q and wait until that thing fully loaded and you get the option screen.
Once your there close the popup and see if your issue is solved.

one-lithe-rune:
@sz1kormar are you able to check whether this is fixed in latest nightly? I put some changes in that should mean it launches in your regular browser in this situation instead of thinking it needs to use MSHTML.

You might need to run with `--no-use_tuned` atm.

PR #1952, opened 2023-11-08
UI: Fix some webui launch corner cases

one-lithe-rune:
Addresses https://github.com/nod-ai/SHARK/issues/1937 amongst other things.

### Motivation

Some days I don't want to have to explain how to use command line options to fix the common and uncommon launch problems people have. And apparently inconsistent branding also irritates me way more than it should - even when that branding might be changing or going away soon.

### Changes

* On windows insist on the presence of webview2 as the embedded browser for `ui=app`. If we can't find it, effectively switch back to `ui=web`. This should prevent pywebview trying to use MSHTML, whilst saying its deprecated, given we are too much for poor old IE11 code trying to run the webui, and it causes a mess. webview2 is pywebviews preferred embedded browser on Window, and Microsoft claims it is installed on all Windows 11, and most Windows 10.
* Add the webview2 runtime droppings to .gitignore. Since I don't use `ui=app` I hadn't noticed them being created when you do that before, as I always force `ui=web`
* If we can't bind to args.server_port get another suitable port from the OS and advise the user that we did this in the UI.
* Make `ui=web` mode use 'SHARK AI Studio' as its title. This makes it consistent with `ui=app`.
* Replace the generic gradio favicon with a nod swirl one instead.

* I haven't tried to uninstall/run without webview2 to get it fall back to MSHTML, and I don't know if its something that's even possible, so I haven't tested that the fallback to `ui=web` actually works.


monorimet (approved):
Thanks, this looks great to me.

PR #1955, opened 2023-11-09
SD/UI Restrict hires fix/img2img resamplers/schedulers

one-lithe-rune:
### Motivation

Hires fix and Image2Image kept blowing up on me because a number of the current resampler options don't exist in PIL which is what the code uses to do the resampling.

Also, since the UI for hires fix uses image2image to do the resize but uses the scheduler you select for text2image, it would blow up again, due to the default there being 'SharkEulerDiscrete' and image2image only working with the CPU schedulers.

This should prevent those problems.

### Changes

* Restrict resamplers for img2img and high res fix to the ones that PIL.Image actually supports, since we uses that to do the resampling. Removed: Antialias, Affine, Cubic. Added: Hamming.
* Set list of available schedulers to CPU only when high res fix is selected in the web ui. Set list to all schedulers when high res fix is deselected.
* Put hi res fix in its own Accordian in the txt2img UI instead of grouping it with Advanced Options.

* The default of 768x768 for hires fix, only works with the 2.1base models at least for me. This doesn't fix that. Anything 1.5 based at 768x768 regardless of app/operation just produces a noise image on my setup.

### DIFF
--- a/apps/stable_diffusion/src/pipelines/pipeline_shark_stable_diffusion_img2img.py
+++ b/apps/stable_diffusion/src/pipelines/pipeline_shark_stable_diffusion_img2img.py
@@ -29,6 +29,10 @@ from apps.stable_diffusion.src.models import (
     SharkifyStableDiffusionModel,
     get_vae_encode,
 )
+from apps.stable_diffusion.src.utils import (
+    resamplers,
+    resampler_list,
+)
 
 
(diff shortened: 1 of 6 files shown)


monorimet:
This is super helpful, thanks. LGTM.

monorimet (approved):

PR #1956, opened 2023-11-09
Remove old fp16 resnet example and add CLI option for device in resnet50 script.

monorimet:
The old resnet50_fp16 example relies on CUDA to produce a fp16 resnet50 module. Removing it to replace with a CPU only method, likely via SHARK-Turbine


monorimet:
Closing as no longer targeting main branch of SHARK.

PR #1960, opened 2023-11-10
Add Falcon-GPTQ Support for 2-way sharding

vivekkhandelwal1:


powderluv:
Maybe we should move falcon to stateless turbine before the sharding.

vivekkhandelwal1:
> Maybe we should move falcon to stateless turbine before the sharding.

PhaneeshB (dismissed):
LGTM

Abhishek-Varma (approved):

Issue #1962, opened 2023-11-10
Failure on linux with winreg and tkinter usage


PhaneeshB:
Hey @one-lithe-rune Thanks for your contribution in #1952 ! 

There's just some missing dependency for winreg and tkinter on top of main branch in SHARK

This is what I have (I'm on a linux system)
command - `python apps/stable_diffusion/web/index.py --share`
error
```
ModuleNotFoundError: No module named 'winreg'
``` 
same thing happens with the tkinter import too.

---
After googling for a bit it seems like we would need pip installable substitutes for these
which will need to be added to [requirements.txt](https://github.com/nod-ai/SHARK/blob/main/requirements.txt)

I see `tk` already a part of the venv, and 
[`winregistry`](https://pypi.org/project/winregistry/) seems to be what can be used?

one-lithe-rune:
Hmmm, I wonder why I'm not seeing that with winreg? 

I though it was built in: https://docs.python.org/3/library/winreg.html

Are you on Linux or Mac? If so, maybe the problem is that I need to move both the imports into the functions that use them so they won't be encountered unless you're on windows. The function that does the registry checking stuff does have a guard clause for win32.

I'll push a commit that does that in the first instance.

PR #1963, opened 2023-11-10
UI: Fix webui launch on non-Windows

one-lithe-rune:
Hopefully addresses #1962

### Motivation

My changes for https://github.com/nod-ai/SHARK/pull/1952 have broken webui startup on Linux (and I assume Mac as well).

### Changes

* Moves the imports of winreg and Tk, into the functions that use them, with winreg behind a guard clause. This should hopefully mean that if you're not on Windows or not using `ui=app` we won't trip over either of these due to them not being there.

* I'm not on Linux or Mac, so I can't test that this actually does the job.
* CI not working right or not covering webui startup biting here...?

### DIFF
--- a/apps/stable_diffusion/web/utils/app.py
+++ b/apps/stable_diffusion/web/utils/app.py
@@ -2,26 +2,25 @@ import os
 import sys
 import webview
 import webview.util
-import winreg
 import socket
 
 from contextlib import closing
 from multiprocessing import Process
-from tkinter import Tk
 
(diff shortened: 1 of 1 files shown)


one-lithe-rune:
Needs someone to test manually on Linux and/or Mac.

monorimet:
Testing this now. Thank you!

monorimet (approved):
Fixes the issue for me on 2 dev machines. thanks.

PR #1971, opened 2023-11-14
Add --no-index to setup_venv.sh runtime pip install.

monorimet:


PhaneeshB (approved):
Thanks for the fix @monorimet! 💯

PR #1972, opened 2023-11-14
UI: Display top tag frequencies and model trained against when a LoRA is selected

one-lithe-rune:
### Motivation

Since I can't remember what most of the LoRAs sitting in my `./models/lora folder` do, given their cryptic filenames, the chances of my coming up with the trigger words for them, are, by now, completely non-existent.

So this tries to give some indication for what they might possibly be for LoRAs in .safetensors format. It does this by querying for training metadata in the format written out by https://github.com/kohya-ss/sd-scripts. If it finds some it them displays the tags used on at least 65% of the images trained, ordered by frequency with their borders colored from green (most common) to red (least common).

It also gives an indication of the model trained against (short hash/filename) if that information is present.

### In Action

ui_display_lora_freq_tags

### Changes

* Adds a function to webui utils to read metadata from .safetensors LoRA files. and do limited parsing of the format written out by the Kohya SS scripts (https://github.com/kohya-ss/sd-scripts) to get tag frequency and trained model information.
* Adds a new common_ui_events.py file for gradio event handlers needed for multiple UI tabs, and adds an event handler for binding to the change event of the LoRA selection boxes, that outputs HTML to display the LoRA tag frequency and model information.
* Adds an HTML gradio control to each of the SD tabs to show the LoRA model name, and most frequently trained tags.
* Bind the change event of the LoRA selection box on each tab to our new event handler, with the output set to the relevant HTML control.
* Display an appropriate text instead of the tags, when no LoRA, or a LoRA that doesn't have the needed metadata information is selected.

* Obviously only works on the koyha-ss style metadata, and I don't know how prevalent that is in real life. It seems to be almost all of the fairly specific LoRAs I have, but is absent from the more general ones.
* I don't handle the case where the metadata describes multiple datasets with overlapping tags particular well, basically the last dataset that had frequency information for that tag wins, rather than it trying to take them all into account properly.
* Green <--> Red is pretty color-blind unfriendly.

### DIFF
--- a/apps/stable_diffusion/web/ui/common_ui_events.py
+++ b/apps/stable_diffusion/web/ui/common_ui_events.py
@@ -1,0 +1,55 @@
+from apps.stable_diffusion.web.ui.utils import (
+    HSLHue,
+    hsl_color,
+    get_lora_metadata,
+)
+
+
+# Answers HTML to show the most frequent tags used when a LoRA was trained,
+# taken from the metadata of its .safetensors file.
+def lora_changed(lora_file):
+    # tag frequency percentage, that gets maximum amount of the staring hue
+    TAG_COLOR_THRESHOLD = 0.55
+    # tag frequency percentage, above which a tag is displayed
+    TAG_DISPLAY_THRESHOLD = 0.65
+    # template for the html used to display a tag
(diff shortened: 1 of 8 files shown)


monorimet (approved):
LGTM! Let me test it before we merge.

PR #1976, opened 2023-11-15
Re-enable SD tunings without matmuls.

monorimet:
It seems there is some lowering pass pipeline change for the matmul dispatches. Since the tunings for matmuls in SD is less performance-impactful than other ops, re-enable tunings for ops other than matmuls.

This bumps inference speed on my 7900xtx back up to ~22it/s

### DIFF
--- a/apps/stable_diffusion/src/utils/stable_args.py
+++ b/apps/stable_diffusion/src/utils/stable_args.py
@@ -323,7 +323,7 @@ p.add_argument(
 
 p.add_argument(
     "--use_tuned",
-    default=False,
+    default=True,
     action=argparse.BooleanOptionalAction,
     help="Download and use the tuned version of the model if available.",
 )
--- a/shark/model_annotation.py
+++ b/shark/model_annotation.py
@@ -117,7 +117,7 @@ def walk_children(
             "mhlo.dot",
             "mhlo.dot_general",
             "mhlo.convolution",
-            "linalg.matmul",
+            # "linalg.matmul",
             "linalg.batch_matmul",
             "linalg.conv_2d_nhwc_hwcf",
             "linalg.generic",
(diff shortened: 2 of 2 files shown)


one-lithe-rune:
This is working for me at about 1-1.5 it/s slower than when the last time tuning was working, and 1.5-2.0 it/s faster than untuned. 

Tested on, 2.1base, Cyberrealistic 3.1 (1.5 based) @ max_length=64, and Cyberrealistic 3.1 with a LoRA at max_length=77, all at 512x512, 7900 XTX.

yzhang93:
> It seems there is some lowering pass pipeline change for the matmul dispatches. Since the tunings for matmuls in SD is less performance-impactful than other ops, re-enable tunings for ops other than matmuls.
> 
> This bumps inference speed on my 7900xtx back up to ~22it/s

Do you know exactly what has changed? Is it the format of lowering config? I don't remember if we've made changes following this PR https://github.com/openxla/iree/pull/15032. Even though it doesn't change the performance much we shouldn't disable the matmul tunings.

monorimet:
> > It seems there is some lowering pass pipeline change for the matmul dispatches. Since the tunings for matmuls in SD is less performance-impactful than other ops, re-enable tunings for ops other than matmuls.
> > This bumps inference speed on my 7900xtx back up to ~22it/s
> 
> Do you know exactly what has changed? Is it the format of lowering config? I don't remember if we've made changes following this PR [openxla/iree#15032](https://github.com/openxla/iree/pull/15032). Even though it doesn't change the performance much we shouldn't disable the matmul tunings.

I don't know yet, it seems like the model annotator is having trouble with matmul_transpose_b dispatches -- it annotates them successfully but iree-compile fails to get a lowering config from the linalg.matmul...

I tried adding some support for linalg.matmul_transpose_b but when the model annotator reads it they are still linalg.matmuls -- does this point to maybe RaiseSpecialOps not preserving the lowering config?

I need to look at the IR more closely and get a minimal reproducer. It's a few cycles I may find time for in the next few days but this is incremental towards re-enabling tunings in general so I hope we can tally matmuls back on in a follow-up patch

yzhang93 (approved):

PR #1981, opened 2023-11-16
Save intermediate values of controlnet

gpetters94:


monorimet (approved):
OK this makes sense. Should it be optional? I'm not familiar with whether the auto1111 api always saves these intermediates or only when specified.

Ideally when we do a drag-and-drop/"send to img2img" of a controlnet-generated image we get the full configuration used to generate it-- this might not actually include these intermediaries in the drag and drop functionality but it would be cool if you got them in the UI when you click on a final result in the output gallery.

one-lithe-rune:
I have no idea about controlnet or what what we're trying to with it here, so this might be a complete nonsense idea:

Is this a circumstance where it might make sense to save in an animated image format (say .webp) and use PIL's `append_images` to stuff them all into the same file?

gpetters94:
@one-lithe-rune This is for matching A1111's API behavior, so ideally we'll do whatever they're doing. That'll be part of the last push to update the API, which should happen on Monday or Tuesday. I like the idea of compressing them into one file, but it'd have to be in a way that doesn't conflict with A1111's API.

@monorimet I'll double check A1111's behavior and if there's a flag that enables this I'll add it in the API parity PR. I've got something resembling the UI changes to show these intermediates in the multicontrolnet branch already, so integrating this into it should be pretty easy.

PR #1983, opened 2023-11-17
API/Docs: Fix incorrect cors arguments listing

one-lithe-rune:
### Motivation

Turns out I didn't update the REST api document for connecting to Joboldcpp with a change I made during development to the naming of the command line argument that sets CORS origins, and didn't notice this because I run the API via a script which uses the updated argument.

### Changes

* Replace `api_cors_origin` in the api/koboldcpp doc, with the correct `api_accept_origin`


monorimet (approved):
LGTM

PR #1984, opened 2023-11-17
Use clean_device_info() by default and don't write .mlir to /tmp/

monorimet:
- saves temp .mlir files to ./shark_tmp/ instead of system /tmp/ (causes issues on linux and some windows envs)
- fixes multigpu cases for vulkan device selection


PhaneeshB (approved):
LGTM!

PR #1986, opened 2023-11-17
Fix an importer issue on Linux

gpetters94:
Fix for a small problem that's been breaking the venv on my system.


monorimet (approved):
OK. Let's pin for now.

PR #1988, opened 2023-11-20
Small UI tweaks for chatbot, fix torchvision requirements

monorimet:
- add torchvision to setup_venv.ps1 -- we need this for the torchvision::nms that is now a dependency of controlnet features.
- Don't have bad flashy orange updates when using the chatbot
- Don't limit the height of the chatbot -- there's mixed opinions and solutions around this one. I think the default (400) is just way too small and LLMs generate plenty enough to justify matching the output.


PhaneeshB (approved):
LGTM

PR #1989, opened 2023-11-21
Create specified dir if needed during save_mlir and fix vulkan device fetching without URI/ID

monorimet:
shark_importer's `save_mlir` fails if the directory doesn't exist. Create it if this is the case.

Soft fail a100 SD CI as image similarity check is expected to fail at the moment.

PR #1991, opened 2023-11-22
Add .mlir to startup shark_tmp cleanup

one-lithe-rune:
## Motivation

d051c3a4a7a7cdf645899e05439ccf782fe37b37 moved generation of intermediate `.mlir` files into `./shark_tmp`. This includes them in the deletion of files we already do for gradio temporary image files on studio start.

## Changes

* Add .mlir to the fiiles that are deleted from `./shark_tmp` when studio is started.
* refactor/rename existing gradio temp file cleanup on startup to be consistent with a general `./shark_tmp` cleanup

### DIFF
--- a/apps/stable_diffusion/web/index.py
+++ b/apps/stable_diffusion/web/index.py
@@ -75,11 +75,11 @@ if __name__ == "__main__":
     # Setup to use shark_tmp for gradio's temporary image files and clear any
     # existing temporary images there if they exist. Then we can import gradio.
     # It has to be in this order or gradio ignores what we've set up.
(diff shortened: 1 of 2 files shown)


monorimet:
Thanks. I hesitated to do the same thinking this might delete llama2 mlir that download from the gs://shark_tank/ but now that I think about it, those should still get saved to the .exe directory or $PWD.

This should keep things nice and clean. Thanks

monorimet (approved):
LGTM (CI waiting on https://github.com/nod-ai/SHARK/pull/1989)

PR #1993, opened 2023-11-23
[vicuna.py] Add option to enable tracing

kuhar:
This makes the program wait for tracy profiler to connect before exiting and flush profiling data after each token.

I don't know how to select the tracy iree-runtime variant programatically -- instead, print an error and exit.


antiagainst (approved):

kuhar:
Rebased

antiagainst (approved):

PR #1904, opened 2023-10-19
Use turbine for gen_sharktank imports.

monorimet:
Needs these changes https://github.com/nod-ai/SHARK-Turbine/pull/102 for a successful run for certain models.

TODO: turbine integration for models that accept multiple inputs.
TODO: complete sets of artifacts for turbine path (needs hash generating for each .mlir)

PR #1996, opened 2023-11-28
add support for external weights

dan-garvey:
@monorimet @aviator19941 I added you guys as reviewers but its not ready yet, some cleanup to do, just as an fyi


monorimet (approved):
LGTM. May revive pytests with https://github.com/nod-ai/SHARK/pull/1904 if we need the extra testing on the model import side.

Next up: SHARK core python API bulldozing and -- anything we want to keep should just go in apps

Ideally we implement a good way to manage the various tests for small APIs utility functions,  and UI features incoming -- this was missing from Studio 1.0 and it hurt quite a bit.

PR #2001, opened 2023-12-02
(Studio) Update gradio and multicontrolnet UI.

monorimet:
Updates gradio pin to 4.7, which might just break everything apparently.

What else does this break? I'd like the new gradio features.


monorimet:
FYI @one-lithe-rune you might have opinions about this change

monorimet:
TODO: fix nod logo, see about controlnet issues.

apps/stable_diffusion/shark_sd.spec
```diff
@@ -19,6 +19,9 @@ a = Analysis(
     win_private_assemblies=False,
     cipher=block_cipher,
     noarchive=False,
+    module_collection_mode={
+        'gradio': 'py',  # Collect gradio package as source .py files
```
dan-garvey, line 23: I dont even want to know how you figured out this was required...

apps/stable_diffusion/src/models/model_wrappers.py
```diff
@@ -109,7 +109,7 @@ def process_vmfb_ir_sdxl(extended_model_name, model_name, device, precision):
     if "vulkan" in device:
         _device = args.iree_vulkan_target_triple
         _device = _device.replace("-", "_")
-        vmfb_path = Path(extended_model_name_for_vmfb + f"_{_device}.vmfb")
+        vmfb_path = Path(extended_model_name_for_vmfb + f"_vulkan.vmfb")
```
dan-garvey, line 112: intentional hardcode?

dan-garvey (approved):

apps/stable_diffusion/src/models/model_wrappers.py
monorimet: for now. 
it should be target triple here but we need to change the naming convention globally

apps/stable_diffusion/shark_sd.spec
monorimet: just wasnt collecting some random source files
tried adding metadata and hidden files, nope, ok just gimme it all then

PR #2015, opened 2023-12-05
(Studio 1.0): Merge LoRA Selection Boxes, add a LoRA Strength parameter, and LoRA alpha calculation fixes

one-lithe-rune:
### Motivation

When I was doing the A1111 API compatibility work, I noticed that we were taking the `use_lora` argument, splitting it into a `use_lora` and `hf_lora_id`, that was reflected in the UI, and then when it came to do the generation we combined these back together to end up with what `use_lora` had originally.

So since gradio now allows dropdowns allow entry of non-list value and has for a while, I embarked changing the UI to remove this silliness. But whilst doing that, and working on the LoRA tags display PR, I ran a bunch of txt2img with LoRA applied, and noticed that some LoRAs just gave noise, and other looked like the were being applied much to strongly compared with the images with a similar strength on Civitai. 

Since I *thought* this was down to the hardcoded strength/alpha value in the code, I decided to add setting the strength in the UI too. 

And then once I did that, I then compared the output to ComfyUI, and though I don't expect to the exact same image in Shark, the LoRAs giving just noise in Shark worked fine there, and Shark still looked like overcooking a lot of the others.

TL;DR lots of stuff in LoRA-land was bad and broken. This hopefully much less broken even though I think things are still fairly ugly.

### Changes (so far)

* Merges LoRA selection in the UI into a single selection, rather than one for LoRAs under ./models and another for Hugging Face Id.
* Add LoRA strength to UI, and pipeline parameters.
* Use LoRA embedded alpha values and up tensor dimension * LoRA strength for final alpha when applying LoRA weights rather than a hardcoded value of 0.75. This seems to be how the Koyha scripts are doing it, and this cleanups the pure noise LoRAs for me and brings the look much more inline with ComfyUI.

### ToDo

- [x] Fix scheduler selection (broken by rebase on current main, I think, although main itself is fine).
- [x] Add command line parameter for LoRA strength.
- [ ] Attempt switch to Diffusers to load the LoRA rather than going through our custom ProcessLoRA function. (may not work, existing codepath that tries this doesn't seem to).
- [x] Include LoRA weight in metadata info and PNG import.
- ~[ ] Manual smoke test of generation with LoRA on each non-SDXL UI tab.~ Needs merge of #2024 for this to be practical
- [x] Allow selection LoRA strength above 1.0. Though this is officially out of range, the previous implementation was effectively setting strength higher than this for some LoRAs, so if you want the same level of effect now for those, you'll need to push higher.

### Problems/Concerns

* LoRA strength is baked into the .vmfb currently, and as such I've reflected that in the .vmfb naming. Having to build a set of .vmfbs when you change the LoRA strength to a value not used before is an obvious pain-point.
* We seem to be running through the codepath updating the model with the LoRA weights even when we're loading up an existing .vmfb. This seems strange, but I'm pretty sure it's not something I've introduced. 
* I'm still not 100% sure I have alpha calculation correct, the results look much much more similar to what ComfyUI is giving me, but...
* Everything is still ending up in ProcessLoRA which is the *exception* path for when one style of Diffusers loading breaks.

### Obvious Future Work, that I won't be doing on this PR

* Multiple-LoRA
* Any changes needed for SDXL LoRA
* Lycoris


one-lithe-rune:
Re-targeted PR to SHARK-1.0 branch.

PR #2025, opened 2023-12-06
Many UI fixes and controlnet impovements

monorimet:
- resizing tweaks so images resized to slider settings
- scheduler list tweaks
- controlnet-specific UI fixes and quality fixes


dan-garvey (dismissed):
only cause its old stuff = P

PR #2027, opened 2023-12-07
Fix stencil pipline to use input image

gpetters94:
We've apparently been doing txt2img with stencils instead of img2img with stencils this entire time.


monorimet (approved):
Thanks.

PR #2034, opened 2023-12-12
Studio2/SD: Use more correct LoRA alpha calculation

one-lithe-rune:
### Motivation

Better to make sure Studio2 has these fixes as we go, rather than port them across at a later date.

### Changes

* Updates ProcessLoRA to use both embedded LoRA alpha, and lora_strength optional parameter (default 1.0) when applying LoRA weights.
* Updates ProcessLoRA to cover more dim cases.
* This brings ProcessLoRA into line with PR #2015 against Studio1

### Possible problems/concerns

* I'm assuming the calls to missing functions is okay for the current state of the project, so I didn't try to sort out the existing call to `get_path_stem` on line 111.

### DIFF
--- a/apps/shark_studio/modules/embeddings.py
+++ b/apps/shark_studio/modules/embeddings.py
@@ -3,34 +3,56 @@ import sys
 import torch
 import json
 import safetensors
+from dataclasses import dataclass
 from safetensors.torch import load_file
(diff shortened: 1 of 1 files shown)


monorimet (approved):
Yes!!! thank you!

PR #2043, opened 2023-12-15
(Studio 1.0) Setup a separate `shark1.venv` for Shark-1.0

one-lithe-rune:
*This is for the SHARK-1.0 branch **only**, and any potential branches from that, do *not* merge into main branch*

### Motivation

I need a working SHARK 1.0 whilst SHARK 2.0 is being constructed, and I still have the draft of #2015 I'm working on for SHARK 1.0.  So I need separate .venvs for `main/studio_sd2` and `SHARK-1.0` branches.

Having a `setup_venv` on the SHARK-1.0 branch that creates/maintains a separately named .venv, means I can just run that and not worry I've broken the .venv for the other branches.

### Changes

* Updates setup_venv.ps1 to create and use ./shark1.venv/
* Updates setup_venv.sh to default to creating ./shark1.venv/

* I haven't tested the `setup_venv.sh` change, someone on Linux/Mac will need to check that.
* I'm assuming we want to keep the main branch .venv as `./shark.venv/`
* I had something much more clever checking git commits and such and deciding which .venv to setup, but then I realised that was pointless, since anything not descended from SHARK-1.0 can just use the current `setup_venv` unaltered.
* Still need to remember to activate the right .venv when switching branches (git hook?)

### DIFF
--- a/setup_venv.ps1
+++ b/setup_venv.ps1
@@ -7,13 +7,13 @@
   It checks the Python version installed and installs any required build
   dependencies into a Python virtual environment.
   If that environment does not exist, it creates it.
-  
+
 .PARAMETER update-src
   git pulls latest version
 
 .PARAMETER force
   removes and recreates venv to force update of all dependencies
-  
+
 .EXAMPLE
   .\setup_venv.ps1 --force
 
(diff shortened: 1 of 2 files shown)


monorimet (approved):
Looks good to me.

PR #2047, opened 2023-12-20
Relocate quantized matmul reassociation flag

monorimet:
This flag should be a model/use-case specific addition, not a default CPU compile flag.

We may need to add it to llama2 if it's required for compilation.


AmosLewis (approved):
LGTM

AmosLewis (approved):

PR #2051, opened 2023-12-30
(Shark 1.0) UI: Upgrade to gradio 4.12.0 and fix breakage

one-lithe-rune:
### Motivation

Came back to SHARK for a bit to generate some images (using 1.0), to discover I somehow had a gradio version which broke UI sizing. Then I also noticed that dependabot wanted an upgrade to 4.11 for `main` branch. Looks like a security fix went into that version.

However latest gradio version pulled down by pip for me is 4.12, and that is also breaking UI sizing, so here's an upgrade to that for Shark 1.0, with a work around for the sizing problem. 

### Changes

* Upgrade Shark 1.0 to gradio 4.12.0
* Add javascript workaround for gradio currently ignoring @media rules in custom CSS. This fixes UI not showing at full width on desktop (>1536px width).
* af4b9c744 Re-enable gallery download buttons as they are now fixed in Gradio.

* Workaround needs javascript shenanigans.
* I've added what I think is the relevant entry in `shark_studio_imports.py` for the new javascript file needed by said shenanigans, but since I've never worked out how to get a sensible .exe build on my machine, so I haven't tested that it is picked up correctly.
* There is another @media rule for >1920px width, and I've added a similar workaround for that, but my screen is not actually that big, so I don't know whether it is doing the right thing.
* I haven't noticed any other gradio induced UI problems, but no guarantee I haven't missed some.
* Of course I put up my first PR in a while just after a new (and likely last for 1.0) stable build and all. 🤣

### DIFF
--- a/apps/stable_diffusion/shark_studio_imports.py
+++ b/apps/stable_diffusion/shark_studio_imports.py
@@ -61,6 +61,7 @@ datas += [
     ("src/utils/resources/opt_flags.json", "resources"),
     ("src/utils/resources/base_model.json", "resources"),
     ("web/ui/css/*", "ui/css"),
+    ("web/ui/js/*", "ui/js"),
     ("web/ui/logos/*", "logos"),
     (
         "../language_models/src/pipelines/minigpt4_utils/configs/*",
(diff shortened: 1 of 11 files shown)


monorimet (approved):
LGTM. Thanks for this.

PR #2052, opened 2023-12-30
(Shark 1.0) SD/UI: Merge Lora Selection Boxes, Add LoRA Strength + LoRA fixes

one-lithe-rune:
*The fixes and alpha calculation changes parts of this are already in main/sd-shark2, this is for Shark 1.0*

### Motivation

I was cleaning up my github branches, and accidentally deleted the branch for #2015 closing the PR. Since that was getting very messy, and although it had some problems it *was* working, this is the rebased, squished on a new PR to SHARK-1.0 and not in draft, version.

The story goes I merging the LoRA selection in the UI into a single control, and noticed a lot of LoRAs I had were completely broken, or being applied weirdly. I added a strength parameter as part of debugging that, and ended up doing a number fixes the calculations based on the Kohya scripts implementation to unbreak images and leaving the output looking a lot more like ComfyUI produces.

### Changes

* Merges LoRA selection in the UI into a single selection, rather than one for LoRAs under ./models and another for Hugging Face Id
* Add LoRA strength to UI and pipeline parameters.
* Add a `--lora_strength` command line argument.
* Bake LoRA strength into .vmfb naming when a LoRA is specified.
* Use LoRA embedded alpha values and (up tensor dimension * LoRA strength) for final alpha when applying LoRA weights rather than a hardcoded value of 0.75
* Adds additional cases to the LoRA weight application that are present for weight application in the Kohya scripts.
* Include lora strength when reading and writing png metadata.
* Allow lora_strength to be set above 1.0 in the UI, so similar effects to the prior (overdriven alpha) implementation can be obtained.

* Strength parameter is baked into the .vmfb, which is not at all user-friendly as it needs to build a new .vmfb whenever you change that, but I think fixing that is one best left for the Turbine version.
* The changes are for the 'diffusers breaks loading the LoRA' exception code pathway, but that seems to be only path that is ever taken. I stalled on trying to work out what was going on there, which is why the previous PR got stuck in draft. I've not even tried to go there for this version.
* It seems to apply the selected LoRA to *something* whenever a .vmfb with the LoRA is loaded, not just when it is generated. This is obviously not correct for a baked in LoRA.

### DIFF
--- a/apps/stable_diffusion/src/models/model_wrappers.py
+++ b/apps/stable_diffusion/src/models/model_wrappers.py
@@ -159,6 +159,7 @@ class SharkifyStableDiffusionModel:
         is_sdxl: bool = False,
         stencils: list[str] = [],
         use_lora: str = "",
+        lora_strength: float = 0.75,
         use_quantize: str = None,
         return_mlir: bool = False,
     ):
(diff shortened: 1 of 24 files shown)


monorimet (approved):
LGTM! I'll test this out as well.

PR #2057, opened 2024-01-05
(Shark 1.0) UI/SD Some UX improvements for SDXL

one-lithe-rune:
### Motivation

I tripped over some things when running running the `juggernautXL_v7Rundiffusion` model from the SDXL tab. This tweaks the UI tab with more immediately functional defaults, and tries to reduce the number of failed attempts at unet base model determination when doing initial .vmfb compilation, eliminating some cases where the wrong base model could be chosen. 

### Changes

* SDXL Tab
  * Filter VAEs in dropdown in the same manner as models.
  * Set default VAE selection to 'madebyollin/sdxl-vae-fp16-fix'.
  * Set default image size to 768x768 to match current Vulkan constraints
* SharkifySDModel Base Unet Model Determination.
  * Alway use the model_to_run as the base model for unet, if that model is a base model defined in base_model.json. Skips looping through all possible base models before that model is defined in the file, whilst attempting to compile for them.
  * Allow SharkSDPipelines to define a 'favor_base_models' classmethod, answering a list of sane base model names for the pipeline. Exclude base models not in that list from compilation attempts when trying to determine a base unet model.
  *  Add a 'favor_base_models' method for both Normal and SDXL Txt2Img Pipelines. Define the method as answering 'None' in the base class, so this additional filtering is not applied for other Pipeline types.
  
* I'm not 100% convinced that 'favor_base_models' needs to be a classmethod, but the existing interaction between the pipelines and SharkifyStableDiffusionModel turned out hairy enough to make my head hurt trying to work out where an instance method would go or be called from.
* Doesn't fix sending parameters from Output Gallery to SDXL tab. This was my other main annoyance beyond the (not fixable for Shark 1.0) compilation memory and time requirements.
* If the wrong base model is already set in variants.json this won't correct it.

### DIFF
--- a/apps/stable_diffusion/src/models/model_wrappers.py
+++ b/apps/stable_diffusion/src/models/model_wrappers.py
@@ -162,6 +162,7 @@ class SharkifyStableDiffusionModel:
         lora_strength: float = 0.75,
         use_quantize: str = None,
         return_mlir: bool = False,
+        favored_base_models=None,
     ):
         self.check_params(max_len, width, height)
         self.max_len = max_len
@@ -191,6 +192,7 @@ class SharkifyStableDiffusionModel:
                 )
 
         self.model_id = model_id if custom_weights == "" else custom_weights
+        self.favored_base_models = favored_base_models
         self.custom_vae = custom_vae
         self.precision = precision
         self.base_vae = use_base_vae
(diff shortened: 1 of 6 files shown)


monorimet (approved):
LGTM.

PR #2060, opened 2024-01-09
Add StreamingLLM support to studio2 chat

monorimet:
- adds in call to gen_external_params for quantized weights generation
- separates weights management from model init
- separates vmfb loading from compilation (switch to vmfbRunner)
- switches api test to use a smaller model


apps/shark_studio/api/utils.py
```diff
@@ -140,22 +140,22 @@ def get_output_value(dev_dict):
...
-        rocm_args = get_iree_rocm_args()
-        iree_flags.extend(rocm_args)
-    if cmd_opts.iree_constant_folding == False:
-        iree_flags.append("--iree-opt-const-expr-hoisting=False")
-        iree_flags.append(
-            "--iree-codegen-linalg-max-constant-fold-elements=9223372036854775807"
-        )
-    if cmd_opts.data_tiling == False:
-        iree_flags.append("--iree-opt-data-tiling=False")
+    # if len(cmd_opts.iree_vulkan_target_triple) > 0:
```
dan-garvey, line 143: dead code?

dan-garvey (changes requested):

dan-garvey:
should we adapt to just use the turbine llm runner?

apps/shark_studio/api/utils.py
monorimet: no, will get used in sd-studio2
we can remove it and readd though

monorimet:
> should we adapt to just use the turbine llm runner?

sounds OK to me, but will need some work to integrate with the UI
might help with maintenance since turbine seems to be the favorite for new features/dev workflows... we just need to have an option to run with SRT and its bleeding edge flags etc

apps/shark_studio/api/utils.py
monorimet: actually, this is better removed completely, model-specific compile flags shouldnt be in a utils file

monorimet:
issue filed for the edge case preventing us from running the api test on a small model with externalized weights https://github.com/openxla/iree/issues/16138

apps/shark_studio/tests/api_test.py
```diff
@@ -7,15 +7,17 @@
 import logging
 import unittest
 from apps.shark_studio.api.llm import LanguageModel
+import gc
 
 
 class LLMAPITest(unittest.TestCase):
-    def testLLMSimple(self):
+    def test01_LLMSmall(self):
```
dan-garvey, line 14: weird naming?

apps/shark_studio/web/ui/chat.py
```diff
@@ -41,9 +45,14 @@ def chat_fn(
     precision,
     download_vmfb,
     config_file,
+    streaming_llm,
     cli=False,
 ):
     global language_model
+    print("Prompt prefix: ", prompt_prefix)
```
dan-garvey, line 52: debug?

apps/shark_studio/web/ui/chat.py
```diff
@@ -63,7 +72,7 @@ def chat_fn(
     prefill_time = 0
     is_first = True
     for text, exec_time in language_model.chat(history):
-        history[-1][-1] = text
+        history[-1][-1] = f"{text}{E_SYS}"
```
dan-garvey, line 75: ideally  f"{text}{E_SYS}" would happen at the api layer no?

dan-garvey (changes requested):

apps/shark_studio/tests/api_test.py
monorimet: forces a certain ordering of tests, but won't be truly necessary until more tests are added.

apps/shark_studio/api/llm.py
```diff
@@ -99,30 +175,57 @@ def __init__(
 
     def compile(self) -> None:
```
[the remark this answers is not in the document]
monorimet: mistake, should just be model arch/api-specific compile flags

apps/shark_studio/web/ui/chat.py
monorimet: yep

apps/shark_studio/web/ui/chat.py
monorimet: I thought the same but struggled to reproduce good results after moving this to api layer. IIRC it had to do with how we yield from the underlying function call.

raikonenfnu:
Is there a way we can specify the vmfb and not compile from scratch?

monorimet:
Would be good to put in an option to set self.vmfb_path instead of renaming the target file. Will update in a few.

monorimet:
Specifying vmfb path etc. will take a genuine CLI option interface or an option in the UI, both of which are significant changes and should come in a follow-up.

@dan-garvey I would prefer not to touch the finely balanced prompt handling until SDXL is finished, and we need this patch to clear the turbine CI.

dan-garvey (approved):
Please file issue for prompt handling, this is how we got where we were in 1.0

monorimet:
> Please file issue for prompt handling, this is how we got where we were in 1.0

https://github.com/nod-ai/SHARK/issues/2073 filed

Issue #2061, opened 2024-01-11
(Shark-1.0) setup_venv.ps1 specifies a torch-mlir version that has dropped off the end of package index


one-lithe-rune:
Getting the following on doing `.\setup_venv.psy`:

```
Looking in indexes: https://pypi.org/simple, https://download.pytorch.org/whl/nightly/cpu
Looking in links: https://llvm.github.io/torch-mlir/package-index/
ERROR: Could not find a version that satisfies the requirement torch-mlir==20231210.* (from versions: 20231213.1051, 20231214.1052, 20231215.1053, 20231216.1054, 20231217.1055, 20231218.1056, 20231219.1057, 20231220.1058, 20231221.1059, 20231222.1060, 20231223.1061, 20231224.1062, 20231225.1063, 20231226.1064, 20231227.1065, 20231228.1066, 20231229.1067, 20231230.1068, 20231231.1069, 20240101.1070, 20240102.1071, 20240103.1072, 20240104.1073, 20240105.1074, 20240106.1075, 20240107.1076, 20240108.1077, 20240109.1078, 20240110.1079, 20240111.1080)
ERROR: No matching distribution found for torch-mlir==20231210.*
Looking in links: https://nod-ai.github.io/SRT/pip-release-links.html
Requirement already satisfied: iree-compiler==20231212.* in c:\develop\shark\shark1.venv\lib\site-packages (20231212.612)
Requirement already satisfied: iree-runtime==20231212.* in c:\develop\shark\shark1.venv\lib\site-packages (20231212.612)Requirement already satisfied: numpy in c:\develop\shark\shark1.venv\lib\site-packages (from iree-compiler==20231212.*) (1.26.2)
Requirement already satisfied: PyYAML in c:\develop\shark\shark1.venv\lib\site-packages (from iree-compiler==20231212.*) (6.0.1)
Building SHARK...
Looking in links: https://llvm.github.io/torch-mlir/package-index/, https://nod-ai.github.io/SRT/pip-release-links.html
Obtaining file:///C:/develop/shark
  Installing build dependencies ... error
  error: subprocess-exited-with-error

  × pip subprocess to install build dependencies did not run successfully.
  │ exit code: 1
  ╰─> [12 lines of output]
      Looking in links: https://llvm.github.io/torch-mlir/package-index/, https://nod-ai.github.io/SRT/pip-release-links.html
      Collecting setuptools>=42
[...]

PR #2062, opened 2024-01-11
(Shark 1.0) Fix setup_venv.ps1 not finding pinned torch-mlir

one-lithe-rune:
Addresses #2061 

### Motivation

The version of torch-mlir pinned by d35288e seems to have dropped off of https://llvm.github.io/torch-mlir/package-index/ so `.\setup-venv.ps1` fails to install it.

### Changes

* Adds a local file equivalent of https://llvm.github.io/torch-mlir/package-index/ containing links to the pinned torch_mlir versions (links taken from the torch-mlir releases github page).
* Use the local file instead of the real package index in setup_venv.ps1.

* I would be entirely unsurprised if this is completely the wrong way to handle this.
* Does change anything for the `setup-venv.sh` script as the original commit didn't touch that.

### DIFF
--- a/package-index-torch-mlir.html
+++ b/package-index-torch-mlir.html
@@ -1,0 +1,14 @@
+<!DOCTYPE html>
+<html><head>
+<meta http-equiv="content-type" content="text/html; charset=UTF-8"></head><body>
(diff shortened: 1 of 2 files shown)


monorimet:
Awesome! Thanks for this.

monorimet:
It may need some tweaking for Linux and/or exe builds but generally a good start.

monorimet (approved):
LGTM!

PR #2066, opened 2024-01-17
(Shark 1.0) SD: Fix inpaint (+others) breakage on sharkification

one-lithe-rune:
### Motivation

Inpainting and (presumably other non txt2img operations) is currently breaking at the point where a `SharkifyStableDiffusionModel` is constructed for the appropriate SD pipeline.

### Changes

* Add the @classmethod decorator to the base implementation of `favored_base_models`. This stops running any SD pipeline that does not have a specific implementation of the method (everything but txt2img and txt2img_sdxl), breaking at the Sharkification stage.


monorimet (approved):
LGTM

PR #2070, opened 2024-01-18
Shark2/SD/UI: Respect ckpt_dir, share and server_port args

one-lithe-rune:
### Motivation 

I've been poking at the SD UI for Shark2 trying to clear some pain points just running the thing, before I try embarking on shuffling all the controls about.

### Changes

* Takes whether to generate a gradio live link from the existing --share command line parameter, rather than hardcoding as True.
* Takes server port from existing --server_port command line parameter, rather than hardcoding as 11911.
* Default --ckpt_dir parameter to '../models'
* Use --ckpt_dir rather than hardcoding ../models as the base directory for checkpoints, vae, and lora, etc
* Add a 'checkpoints' directory below --ckpt_dir to match ComfyUI folder structure. Read custom_weights choices in the UI from there, and/or subfolders below there matching the selected base model.
* Fix --ckpt_dir possibly not working correctly when an absolute rather than relative path is specified.
* Relabel "Custom Weights" to "Custom Weights Checkpoint" in the UI

* Side-effect of this is to have the `vmfb` folder ends being made below `models` (or whatever you've specified as --ckpt_dir). Possibly not ideal, but the easiest thing to do.
* Not sure the pipeline will be able to find things correctly, now I've twiddled with things, as the pipelines are broken for me due to errors involving unexpected custom_vae keyword arguments, and Turbine not support compiling MLIR on windows.

### DIFF
--- a/apps/shark_studio/modules/shared_cmd_opts.py
+++ b/apps/shark_studio/modules/shared_cmd_opts.py
@@ -601,7 +601,7 @@ p.add_argument(
 p.add_argument(
     "--ckpt_dir",
     type=str,
-    default="",
+    default="../models",
     help="Path to directory where all .ckpts are stored in order to populate "
     "them in the web UI.",
 )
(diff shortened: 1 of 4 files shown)


monorimet:
LGTM! Thanks.

PR #2087, opened 2024-02-08
Update README to recommend SHARK-1.0 for now

one-lithe-rune:
## Motivation

When people just want an image generator its too easy for them to grab a not yet working Shark2 version from `main`, rather than getting SHARK-1.0.

This updates the README.md to tell them they should be using SHARK-1.0 for now.

## Changes

- Add a note at the of the top of the README to use SHARK-1.0 whilst rewrite with Turbine is ongoing.
- Update installation section to point at SHARK-1.0 branch.
- Suggest the latest SHARK-1.0 pre-release as well as stable.
- Recommend running the .exe from the command line.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,18 +2,20 @@
 
 High Performance Machine Learning Distribution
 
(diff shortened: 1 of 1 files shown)


monorimet (approved):
LGTM.

PR #2088, opened 2024-02-09
Add Controlnet to 2.0

gpetters94:


apps/shark_studio/web/ui/sd.py
```diff
@@ -731,7 +745,7 @@ def base_model_changed(base_model_id):
         neg_prompt_submit = negative_prompt.submit(**status_kwargs).then(**pull_kwargs)
         generate_click = (
             stable_diffusion.click(**status_kwargs)
-            .then(**pull_kwargs)
+            # .then(**pull_kwargs)
```
monorimet, line 748: it's been a while since I put this together. does this need to be commented out for some reason? we can revise the "import to config" behavior if needed.

PR #2089, opened 2024-02-16
Add *.safetensors to .gitignore

one-lithe-rune:
### Motivation

Whilst trying to remove some more blockages to Shark2 generating an image on Windows, I kept ending up with .safetensors files appearing and `git` still noticing them.

### Changes

- shark2 is putting base model .safetensors files in model specific subfolders. Easiest to just ignore .safetensors completely.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -183,6 +183,7 @@ generated_imgs/
 # Custom model related artefacts
 variants.json
 /models/
+*.safetensors
 
 # models folder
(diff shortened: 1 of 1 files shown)


monorimet (approved):

monorimet:
Ideally all the .safetensors will just be saved to the user's HF cache if it exists. This will help in the case where it doesn't and until that's actually a feature. Thanks!

PR #2091, opened 2024-02-18
Studio2/SD/UI: Further sd ui pipeline fixes

one-lithe-rune:
### Motivation

On Windows, this gets us all the way to failing in iree compile of SD 2.1 base; per @monorimet this is where we expect it to fail with the current Turbine build.

You will need to `.\setup_venv.ps1 --force` to clear out the conflicting dependencies previously sourced from SRT etc.

### Updates

- Fix merge errors with sd right pane config UI tab.
- Remove non-requirement.txt install/build of torch/mlir/iree/SRT in setup_venv.ps1, fixing "torch.compile not supported on Windows" error.
- Fix gradio deprecation warning for `root=` FileExplorer kwarg.
- Comment out `precision` and `max_length` kwargs being passed to unet, as not yet supported on main Turbine branch. Avoids keyword argument error.

- As usual I haven't made any changes to `./setup_venv.sh`, as I'm not currently in a position to test that. It may need equivalent changes to `.\setup_venv.ps1`.
- My editor is apparently not auto-wrapping commit messages anymore. 🤦

### DIFF
--- a/apps/shark_studio/api/sd.py
+++ b/apps/shark_studio/api/sd.py
@@ -120,8 +120,8 @@ class StableDiffusion(SharkPipelineBase):
                 # "num_loras": num_loras,
                 "height": height,
                 "width": width,
-                "precision": precision,
-                "max_length": self.model_max_length,
+                # "precision": precision,
+                # "max_length": self.model_max_length,
             },
             "vae_encode": {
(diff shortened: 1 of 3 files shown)


monorimet (approved):
LGTM!

PR #2097, opened 2024-02-25
Studio2/SD/UI: Update gradio to 4.19.2 (sd-studio2)

one-lithe-rune:
### Motivation

Dependabot wants gradio to go to 4.19.2. This does that for `sd-studio2`.

### Changes

- Move pin for gradio from 4.15 -> 4.19.2 on the sd-studio2 branch

- There don't appear to be any changes needed other than the version bump. So maybe it would make more sense to just merge back from `main` if that gets updated?

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -26,7 +26,7 @@ parameterized
 accelerate
 scipy
 ftfy
-gradio==4.15.0
+gradio==4.19.2
 altair
 omegaconf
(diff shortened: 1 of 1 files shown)


monorimet:
@gpetters-amd does this work with what you have going on the .exe

monorimet (approved):
Works in the .exe for me.
Thanks!

PR #2098, opened 2024-02-26
Shark-1.0: Update gradio to 4.19.2

one-lithe-rune:
### Motivation

Dependabot wants an update of `gradio` to 4.19.2. This changed the behaviour of gradio change events slightly and revealed a bug that previously didn't trigger. That bug caused a crash when doing .png import for txt2img and sdxl txt2imge, or sending to either of those tabs.

### Changes

- Move pin for gradio from 4.12 -> 4.19.2
- Fix png import txt2img/txt2img_sdxl event handling for gradio 4.19.2

- black also wanted a bunch of reformatting of existing code, because new black version + new year. So a lot of files not involved in the gradio update are also changed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -168,6 +168,7 @@ cython_debug/
 *venv/
 shark_tmp/
 *.vmfb
+*.safetensors
 .use-iree
 tank/dict_configs.py
 *.csv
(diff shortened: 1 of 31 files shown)


monorimet (approved):
LGTM, thanks for this.

Issue #2095, opened 2024-02-23
"AttributeError: module 'mpmath' has no attribute 'rational'" when recreating .venv (+workaround)


one-lithe-rune:
Found this when trying to setup the .venv on Linux. But this happens for me on both `SHARK-1.0` and `sd-studio2` branches, on both Windows and Linux.

The workaround is to do `pip install mpmath==1.3.0`. But I'm guessing its an upstream requirements problem rather than a SHARK requirements one?

```
(shark1.venv) PS C:\develop\SHARK> shark-run
72175701 (HEAD -> SHARK-1.0, tag: 20240126.1139, upstream/SHARK-1.0, origin/SHARK-1.0) SDXL tweaks (#2072)
On branch SHARK-1.0
Your branch is up to date with 'upstream/SHARK-1.0'.

Untracked files:
  (use "git add <file>..." to include in what will be committed)
        apps/shark_studio/web/models/

nothing added to commit but untracked files present (use "git add" to track)
Traceback (most recent call last):
  File "C:\develop\SHARK\apps\stable_diffusion\web\index.py", line 5, in <module>
    import apps.stable_diffusion.web.utils.app as app
  File "C:\develop\SHARK\apps\stable_diffusion\web\utils\app.py", line 10, in <module>
    from apps.stable_diffusion.src import args
  File "C:\develop\SHARK\apps\stable_diffusion\src\__init__.py", line 1, in <module>
    from apps.stable_diffusion.src.utils import (
  File "C:\develop\SHARK\apps\stable_diffusion\src\utils\__init__.py", line 12, in <module>
    from apps.stable_diffusion.src.utils.sd_annotation import sd_model_annotation
  File "C:\develop\SHARK\apps\stable_diffusion\src\utils\sd_annotation.py", line 3, in <module>
    from shark.model_annotation import model_annotation, create_context
  File "C:\develop\SHARK\shark\__init__.py", line 4, in <module>
    from torch._dynamo import register_backend
  File "C:\develop\SHARK\shark1.venv\Lib\site-packages\torch\_dynamo\__init__.py", line 2, in <module>
    from . import allowed_functions, convert_frame, eval_frame, resume_execution
  File "C:\develop\SHARK\shark1.venv\Lib\site-packages\torch\_dynamo\allowed_functions.py", line 30, in <module>
    from .utils import hashable, is_safe_constant, NP_SUPPORTED_MODULES
[...]

PR #2099, opened 2024-02-27
SHARK-1.0: Fixes to requirements for newer dependency versions

one-lithe-rune:
### Motivation

Setting up a `venv` for SHARK-1.0 was getting to be a right pain, due to #2095 and there apparently no longer being any common torch version that fits both torchvision and the version of torch-mlir we have pinned for SHARK-1.0.

### Changes

- Adds mpmath==1.3.0 as a direct requirement to workaround a sympy dependency issue.
- Change setup_venv.ps1 to install torchvision --nodeps. This avoids a failure to find any common torch version dependency between the pinned torch-mlir version and torchvision after a long search. This also more closely matches what setup_venv.sh does.

- Still complains about the dependency conflict around torchvision, but at least it doesn't now check every listed version of torchvision at the index url before giving up.
- Shouldn't we just `pip --freeze` everything and pin it all for Shark-1.0,given the current dev effort is all on Studio2/Turbine?

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -45,6 +45,9 @@ langchain
 einops # for zoedepth
 pydantic==2.4.1 # pin until pyinstaller-hooks-contrib works with beta versions
 
+# mpmath sub-dependency, breaks sympy sub-dependency as of (1.4.0a0) version
+mpmath==1.3.0
+
 # Keep PyInstaller at the end. Sometimes Windows Defender flags it but most folks can continue even if it errors
 pefile
 pyinstaller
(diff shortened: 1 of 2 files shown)


monorimet (approved):
Thanks, LGTM. I haven't been looking at 1.0 but I can't think of any reason not to pin everything.

ameeuwsen:
I'm still getting some wonkiness around this area. On `./setup_venv.ps1` everything seems to run well except when it comes to `pip install torchvision`. There seems to be another version clashing. It downgrades `torch 2.40.` to `torch 2.2.0.devxxxx`. Error below.

`ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
torch-mlir 20231210.1048 requires torch==2.2.0.dev20231204, but you have torch 2.4.0 which is incompatible.`

PR #2103, opened 2024-03-07
Update _IREE_TARGET_MAP to return 'vulkan-spirv' for 'vulkan'

one-lithe-rune:
### Motivation

After updating the .venv, `iree` started complaining that that `vulkan` was a not a valid output backend.

### Changes

- Change target passed to iree for vulkan from 'vulkan' to 'vulkan-spriv', as 'vulkan' is not a valid value for --iree-hal-target-backends with the current iree compiler.

- https://github.com/openxla/iree/pull/16591 implies that using --iree-hal-target-backends is not going to be the right approach going forward, but it is what `ireec.compilestr` uses, which is what Shark is calling in turn 🤷


monorimet (approved):
Thanks! LGTM

PR #2114, opened 2024-04-08
Temporarily remove target env

gpetters-amd:


monorimet (dismissed):
I think we can remove it entirely as these should really be upstreamable.

monorimet (approved):
LGTM!

PR #2127, opened 2024-04-23
Clean up the filesystem stuff and add support for HF ckpt downloading

gpetters-amd:
This works fine for SD2.1 but I've noticed an issue downloading the safetensors file for clip for SD1.5. There's still some more cleanup necessary here, so I'd appreciate a second pair of eyes on it. I'm also not sure if the way I'm centralizing all the filesystem logic (just using the args and calling `os.path.join` at each usage) is the right thing to do or if we want to have everything go through, say, functions in `file_utils.py`.

### DIFF
--- a/apps/shark_studio/web/ui/sd.py
+++ b/apps/shark_studio/web/ui/sd.py
@@ -231,9 +231,10 @@ def import_original(original_img, width, height):
 
 
 def base_model_changed(base_model_id):
-    new_choices = get_checkpoints(
-        os.path.join("checkpoints", os.path.basename(str(base_model_id)))
-    ) + get_checkpoints(model_type="checkpoints")
+    ckpt_path = Path(os.path.join(cmd_opts.model_dir, "checkpoints", os.path.basename(str(base_model_id))))
+    ckpt_path.mkdir(parents=True, exist_ok=True)
+
+    new_choices = get_checkpoints(ckpt_path) + get_checkpoints(model_type="checkpoints")
 
     return gr.Dropdown(
         value=new_choices[0] if len(new_choices) > 0 else "None",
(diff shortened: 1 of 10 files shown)


apps/shark_studio/web/ui/sd.py
```diff
@@ -392,6 +397,9 @@ def base_model_changed(base_model_id):
                         elem_id="negative_prompt_box",
                         show_copy_button=True,
                     )
+                    with gr.Row():
```
monorimet, line 400: Why do we add an explicit option to save the prompt? It should write prompts to the json config, so this seems a bit redundant to me.

apps/shark_studio/web/ui/sd.py
```diff
@@ -631,19 +624,18 @@ def base_model_changed(base_model_id):
...
+                            default_config_file = os.path.join(
+                                get_configs_path(),
+                                "default_sd_config.json",
+                            )
+                            write_default_sd_config(default_config_file)
+                            sd_json = gr.JSON(
+                                label="SD Config",
+                                elem_classes=["fill"],
+                                value=view_json_file(default_config_file),
+                                render=False,
```
monorimet, line 637: Why turn off rendering this?

monorimet:
Thanks @gpetters-amd. I left a few comments, but looks better for the most part.

apps/shark_studio/web/ui/sd.py
gpetters-amd: It's the swap the order - I moved the render call to line 702 below.

apps/shark_studio/web/ui/sd.py
gpetters-amd: Yeah, the new config stuff is half-baked. Took it out, we can discuss how to handle it after we've got stable out.

monorimet (approved):
Looks better. Thank you.

PR #2153, opened 2024-05-31
Fixes to UI config defaults, config loading, and warnings.

monorimet:


apps/shark_studio/api/sd.py
```diff
@@ -285,6 +286,32 @@ def shark_sd_fn_dict_input(
         if key == "seed":
             sd_kwargs[key] = int(sd_kwargs[key])
 
+    # TODO: move these checks into the UI code so we don't have gradio warnings in a generalized dict input function.
+    if sd_kwargs["device"] == "":
```
gpetters94, line 290: Nit: if this can be None it might be better if the check is `if not sd_kwargs["device"]`

gpetters94 (approved):
LGTM

gpetters94 (approved):

PR #2170, opened 2024-09-19
Change Llama2 from the Turbine implementation to the Sharktank one

gpetters-amd:
There are still two outstanding issues I'd like some comments on, but otherwise this should be basically done.

### DIFF
--- a/apps/shark_studio/api/llm.py
+++ b/apps/shark_studio/api/llm.py
@@ -1,4 +1,9 @@
-from turbine_models.custom_models import stateless_llama
+from shark_turbine.aot import *
+from sharktank.models.llama.llama import LlamaModelConfig, PagedLlamaModelV1
+import sharktank
+import huggingface_hub
+
+# from turbine_models.custom_models import stateless_llama
 from turbine_models.model_runner import vmfbRunner
 from turbine_models.gen_external_params.gen_external_params import gen_external_params
 import time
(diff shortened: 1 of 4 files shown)


apps/shark_studio/api/llm.py
```diff
@@ -130,13 +135,18 @@ def __init__(
...
+                #    quantization=self.quantization,
+                #    weight_path=self.external_weight_file,
+                #    hf_auth_token=hf_auth_token,
+                #    precision=self.precision,
+                # )
+                cache_dir = os.path.join(".", str(self.hf_model_name).replace("/", "_"))
+                huggingface_hub.snapshot_download(
+                    repo_id=self.hf_model_name, cache_dir=cache_dir
                 )
+                # TODO: Convert to gguf, delete cache
```
gpetters-amd, line 149: The way that sharktank recommends for generating the .gguf file is to use a CLI tool from llama.cpp. Is that still the best way to extract that, or do we have a way to do it using sharktank?

apps/shark_studio/api/llm.py
```diff
@@ -161,20 +171,39 @@ def __init__(
...
+            llama_config = sharktank.models.llama.llama.LlamaModelConfig(hp)
+            llama_config.use_hf = False
+            llama_config.static_tables = (
+                False  # Rely on the compiler for hoisting tables.
+            )
+            llama_config.kv_cache_type = "direct"  # if args.bs == [1] else "paged"
+            model = PagedLlamaModelV1(dataset.root_theta, llama_config)
+
+            fxb = FxProgramsBuilder(model)
+            self.torch_ir = export(fxb)
```
gpetters-amd, line 202: Not sure why, but this is producing an empty module. Any idea what I'm missing?

PR #2171, opened 2024-09-24
SHARK API

gpetters-amd:
This is the first step toward the SHARK refactor, right now it's just a skeleton but once I'm sure this is what we're going for I can add in the model complilation code (that's mostly going to be ported from Turbine) and get SD2 working soon.

### DIFF
--- a/apps/shark_studio/api/shark_api.py
+++ b/apps/shark_studio/api/shark_api.py
@@ -1,0 +1,200 @@
+
+# Internal API
+
+# Used for filenames as well as the key for the global cache
+def safe_name():
+    pass
+
+def local_path():
+    pass
+
+def generate_sd_vmfb(
+    model: str,
+    height: int,
+    width: int,
+    steps: int,
+    strength: float,
+    guidance_scale: float,
+    batch_size: int = 1,
+    base_model_id: str,
+    precision: str,
+    controlled: bool,
+    **kwargs,
+):
+    pass
+
+def load_sd_vmfb(
+    model: str,
+    weight_file: str,
+    height: int,
+    width: int,
+    steps: int,
+    strength: float,
+    guidance_scale: float,
+    batch_size: int = 1,
+    base_model: str,
+    precision: str,
+    controlled: bool,
+    try_download: bool,
+    **kwargs,
+):
+    # Check if the file is already loaded and cached
+    # Check if the file already exists on disk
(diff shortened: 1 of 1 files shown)


apps/shark_studio/api/shark_api.py
```diff
@@ -0,0 +1,200 @@
...
+    try_download: bool,
+    **kwargs,
+):
+    # Check if the file is already loaded and cached
+    # Check if the file already exists on disk
+    # Try to download from the web
+    # Generate the vmfb (generate_sd_vmfb)
+    # Load the vmfb and weights
+    # Return wrapper
+    pass
```
monorimet, line 47: This conflicts a bit with what the actual engine (be it a turbine-models pipeline or the next thing) will be responsible for.

Think of, for example, a huggingface pipeline. You instantiate it with one line, you send it to GPU, and you run it.

For us, you prepare the pipeline (instantiate, prepare vmfbs), load it, and run. No need to worry about "vmfb" from the webui's perspective.

Whatever engine we have should be able to distill into "configure / load" as above, perhaps with an outer layer of queueing in datacenter cases. And in that case, the engine will take care of it and expose a fastapi endpoint or the like. Studio should just send requests to a server and display the results.

That being said, some of these signatures can be simplified.

"init" should be generalized to "I need to init if I am running a different engine" -- this means height, width, model id, anything that changes the IR of one or more of the large components of the job (clip, unet, vae).

Schedulers being the exception (they are small enough to export JIT), the bulk of export/compile should happen at init.

Weights configuration and loading at load time.

guidance scale, init image, prompts, runtime-exposed knobs go to generate_images. Right now this includes JIT scheduler prepare/load since it takes only a second or two, but really they should just be fully dynamic and prepared at init/load time. For the API it just needs num_inference_steps and scheduler_id at `generate_images` to accomplish this.

gpetters-amd: Gotcha. Wasn't sure if we were moving away from pipelining or not, shouldn't be a huge fix.

monorimet:
@gpetters-amd FWIW the existing api file in apps/shark_studio/api/sd.py can be converted to apps/shark_studio/api/turbine-models-sd.py and refactored to match an abstracted API established in this PR. That way we don't lose the idea you had of an inference engine-agnostic interface to the Studio WebUI.

